PR #9, opened 2022-05-27
Ddc client

upalinski:


packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -86,18 +93,20 @@ export class ContentAddressableStorage {
             );
         }
 
-        return this.toPiece(pbSignedPiece.piece);
+        return this.toPiece(pbSignedPiece.piece, cid);
     }
 
     async search(query: Query): Promise<SearchResult> {
+        //ToDo add piece CID to result
```
upalinski, line 100: is it outdated TODO item?

packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -8,27 +8,34 @@ import {Piece} from "./models/Piece";
...
+    readonly cdnNodeUrl: string;
 
-    cidBuilder: CidBuilder;
+    readonly cipher?: CipherInterface;
+    readonly cidBuilder: CidBuilder;
 
     constructor(
         scheme: SchemeInterface,
-        gatewayNodeUrl: string,
+        cdnNodeUrl: string,
```
upalinski, line 31: client supposed to be configured by cdn cluster id and use smart contract to discover cluster nodes (fallback on other node in case some is unavailable)

packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -112,17 +121,31 @@ export class ContentAddressableStorage {
...
+            .filter(p => isPiece(p.signedPiece?.piece))
+            .map(e => this.toPiece(e.signedPiece!.piece!, e.cid))
 
         return new SearchResult(pieces);
     }
 
-    private toPiece(piece: PbPiece): Piece {
+    async storeEncrypted(bucketId: bigint, piece: Piece, encryptionOptions: EncryptionOptions): Promise<PieceUri> {
+        const encryptedPiece = piece.clone();
+        encryptedPiece.tags.push(new Tag(ENCRYPTED_KEY, "true"));
```
upalinski, line 133: i would recommend to change constants postfix from KEY to TAG. So it would be like:

```
encryptedPiece.tags.push(new Tag(ENCRYPTED_TAG, "true"));
encryptedPiece.tags.push(new Tag(DEK_PATH_TAG, encryptionOptions.dekPath));
```

What do you think?

packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -112,17 +121,31 @@ export class ContentAddressableStorage {
...
+        encryptedPiece.tags.push(new Tag(DEK_PATH_KEY, encryptionOptions.dekPath));
+        encryptedPiece.data = this.cipher!.encrypt(piece.data, encryptionOptions.dek);
+        return this.store(bucketId, encryptedPiece)
+    }
+
+    async readDecrypted(bucketId: bigint, cid: string, dek: Uint8Array): Promise<Piece> {
```
upalinski, line 139: we could have single read method and check for existence of 'ENCRYPTED' tag. Just proposed an option, separate methods are also fine for me

packages/core/src/crypto/signature/Scheme.ts
```diff
@@ -15,15 +17,15 @@ export class Scheme implements SchemeInterface{
         this.publicKeyHex = publicKeyHex;
     }
 
-    static async createScheme(scheme: "sr25519" | "ed25519", seedHex: string): Promise<Scheme> {
+    static async createScheme(scheme: SchemeType, seedHex: string): Promise<Scheme> {
```
upalinski, line 20: rename seedHex -> mnemonic?

packages/ddc-client/src/DdcClient.ts
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+
+    async read(pieceUri: PieceUri, options: ReadOptions = {}): Promise<PieceArray> {
+        const headPiece = await this.caStorage.read(pieceUri.bucketId, pieceUri.cid);
+        const isEncrypted = headPiece.tags.filter(t => t.key == ENCRYPTED_KEY && t.value == "true").length > 0;
+
+        //TODO 4. put into DEK cache
+        let objectDek = new Uint8Array();
+        if (options.decrypt) {
```
upalinski, line 156: do we need this flag if we have 'isEncrypted' tag in a piece?

packages/file-storage/src/core/CoreFileStorage.ts
```diff
@@ -31,24 +25,25 @@ export class CoreFileStorage {
         const links = indexedLinks
             .sort((a, b) => a.position - b.position)
             .map(e => e.link);
-        return await this.caStorage.store(bucketId, new Piece(encoder.encode("metadata"), [], links));
+
+        const piece = new Piece(encoder.encode("metadata"), tags, links);
```
naure, line 29: What is this constant "metadata" ? I think we should leave the data field empty.

packages/ddc-client/src/DdcClient.ts
```diff
@@ -0,0 +1,234 @@
...
+            } else if (!dekPath.startsWith(options.dekPath! + "/") && dekPath !== options.dekPath!) {
+                throw new Error(`Provided dekPath='${options.dekPath}' doesn't correct for piece with dekPath='${dekPath}'`);
+            }
+
+            const clientDek = await this.readDek(pieceUri.bucketId, options.dekPath!);
+
+            objectDek = DdcClient.buildHierarchicalDekHex(clientDek, dekPath.replace(options.dekPath!, "").replace(/^\//, ""))
+        }
+
+        if (headPiece.links.length > 0) {
```
upalinski, line 169: as option we can use separate tag 'multipart' to detect multipart (chuncked) files

packages/ddc-client/src/model/PieceArray.ts
```diff
@@ -0,0 +1,37 @@
+import {Tag} from "@cere-ddc-sdk/content-addressable-storage";
+
+type Data = ReadableStream<Uint8Array> | string | Uint8Array
+
+export class PieceArray {
```
upalinski, line 5: might be just Piece? And isPiece method can be renamed to isSinglePart or use !isMultipart

packages/ddc-client/src/options/ClientOptions.ts
```diff
@@ -0,0 +1,37 @@
+//TODO put comments
+import {CidBuilder, CipherInterface, NaclCipher, SchemeInterface, SchemeType} from "@cere-ddc-sdk/core";
+import {Options as SmartContractOptions, TESTNET} from "@cere-ddc-sdk/smart-contract";
+import {MB} from "@cere-ddc-sdk/file-storage";
+
+export class ClientOptions {
+     clusterAddress: string | number; // Cluster ID or CDN URL
```
upalinski, line 7: do we need this 'Or'? Can we replace by clusterId?

packages/core/src/crypto/signature/Scheme.ts
naure: We’re trying to get away from "mnemonic", preferring _secret phrase_ for the human-readable words and _seed_ for the derived bytes.

upalinski: oh, yes. Forgot about it. But still it should be renamed from `seedHex` to `secret phrase`

packages/content-addressable-storage/src/ContentAddressableStorage.ts
naure: Or simply use the presence of a DEK_PATH tag to mean that it is encrypted.

packages/ddc-client/src/options/ClientOptions.ts
naure: It makes sense to support two options: find URLs from cluster, or use a specific URL given by the app.

packages/ddc-client/src/DdcClient.ts
upalinski: i think about the future. Later links can be used not only for the 'chunked' pieces but to link pieces (relation) so suggesting to don't use a presence of links as identifier is it multipart file or not and use a separate tag 'multipart':'true' in the main descriptor piece.

packages/content-addressable-storage/src/ContentAddressableStorage.ts
svalasovich: removed

packages/file-storage/src/core/CoreFileStorage.ts
svalasovich: removed

packages/ddc-client/src/options/ClientOptions.ts
svalasovich: Yes, I think have feature by using only CDN url is useful

packages/content-addressable-storage/src/ContentAddressableStorage.ts
svalasovich: renamed and removed ENCRYPTED_TAG

packages/ddc-client/src/model/PieceArray.ts
svalasovich: Need to discuss about object naming. Method renamed to `isMultipart`

packages/ddc-client/src/DdcClient.ts
svalasovich: I think we need, because there are chance, that user doesn't want to decrypt encrypted message. It can be some ETL job which transfer data from one cluster to other

packages/content-addressable-storage/src/ContentAddressableStorage.ts
svalasovich: added discovery by cluster id for ca storage

packages/ddc-client/src/DdcClient.ts
upalinski: makes sense. ok

PR #30, opened 2022-06-29
Schemas v0.1.3 - Test vector

naure:


packages/content-addressable-storage/src/ContentAddressableStorage.spec.ts
```diff
@@ -0,0 +1,58 @@
+import {ContentAddressableStorage, Piece} from "./index.js"
+import {u8aToHex, hexToU8a} from "@polkadot/util";
+import * as fs from 'fs/promises';
+import * as path from 'path';
+
+// To generate a new version of the test vector, change the filename, re-run
+// the test, and commit the new value in the ddc-schemas repo (git submodule).
+const VECTOR_FILE = "ddc-schemas/test-vectors/store-request-sdk-js-1.2.8.json";
```
svalasovich, line 8: Could you write flow? We build json file with requests in ddc-schemas, then execute prepared requests in E2E tests? That is idea?

svalasovich (approved):

packages/content-addressable-storage/src/ContentAddressableStorage.spec.ts
naure: You can use this in different places. I think it’s most useful in unittests.

> …  They can be used to test for regressions or compatibility between different components at different versions. …

svalasovich (approved):

PR #31, opened 2022-06-29
Protobuf from DDC Schemas and sign structure update

svalasovich:


packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -77,12 +77,10 @@ export class ContentAddressableStorage {
 
     async store(bucketId: bigint, piece: Piece): Promise<PieceUri> {
         const pbPiece: PbPiece = {
-            bucketId: bucketId.toString(),
+            bucketId: Number(bucketId),
```
naure, line 80: You could drop completely the idea of bigint. That’s the advantage of using uint32.

svalasovich: In future I will. I don't want to change now, because it will affect docs and other users code from 1.2.8 -> 1.2.9

PR #35, opened 2022-07-01
Bucket alloc

svalasovich:


packages/ddc-client/src/DdcClient.ts
```diff
@@ -93,7 +93,10 @@ export class DdcClient implements DdcClientInterface {
         if (balance > 0) {
             await this.smartContract.accountDeposit(balance);
         }
-        await this.smartContract.bucketAllocIntoCluster(event.bucketId, resource);
+
+        const clusterStatus = await this.smartContract.clusterGet(Number(clusterId));
+        const bucketSize = BigInt((Number(resource * 1000n) / clusterStatus.cluster.vnodes.length) | 0);
```
upalinski, line 98: so, then minimal size of the bucket is 1 GB, right? @naure are we ok with that?

upalinski, line 98: i think formula isn't correct. As we discussed before, it should be next:
bucket_size / vnodes.length * replication factor

in your code you don't use replication factor 

@naure please, confirm that formula above is correct

packages/ddc-client/src/DdcClient.ts
```diff
@@ -104,13 +107,15 @@ export class DdcClient implements DdcClientInterface {
 
     async bucketAllocIntoCluster(bucketId: bigint, resource: bigint) {
         const bucketStatus = await this.bucketGet(bucketId);
+        const clusterStatus = await this.smartContract.clusterGet(bucketStatus.bucket.cluster_id);
 
-        const total = BigInt(bucketStatus.bucket.resource_reserved) + resource;
+        const total = BigInt(bucketStatus.bucket.resource_reserved * clusterStatus.cluster.vnodes.length) / 1000n + resource;
```
upalinski, line 112: if formula i've provided above is correct, than you should use bucket replication factor here

packages/ddc-client/src/DdcClient.ts
```diff
@@ -104,13 +107,15 @@ export class DdcClient implements DdcClientInterface {
...
         if (total > MAX_BUCKET_SIZE) {
             throw new Error(`Exceed bucket size. Should be less than ${MAX_BUCKET_SIZE}`);
         }
 
-        await this.smartContract.bucketAllocIntoCluster(bucketId, resource);
+        const bucketSize = BigInt((Number(resource * 1000n) / clusterStatus.cluster.vnodes.length) | 0);
```
upalinski, line 117: it's not a bucket size, might be `resourcesToAllocate`?

svalasovich: Renamed

PR #37, opened 2022-07-26
Improve packages creation

aelyseev:


packages/core/package.json
```diff
@@ -8,24 +8,55 @@
...
   },
   "devDependencies": {
-    "typescript": "4.5.4"
-  },
-  "peerDependencies": {
-    "@polkadot/keyring": "^6.0.5",
-    "@polkadot/util": "^6.0.5",
-    "@polkadot/util-crypto": "^6.0.5",
-    "@polkadot/wasm-crypto": "^4.0.2"
+    "@multiformats/blake2": "^1.0.10",
```
svalasovich, line 53: We have many packages like this, where sdk uses them in code. Don't we need to move them to `dependencies` instead `devDependencies`

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cere-ddc-sdk",
   "description": "Cere DDC SDK for JavaScript",
-  "version": "1.2.9",
+  "version": "1.5.40",
```
svalasovich, line 4: Set version to 1.2.9. We will update as separate PR before publishing

packages/core/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+import { createRequire } from "module";
+
+const require = createRequire(import.meta.url);
+const browserslistEnv = process.env.BROWSERSLIST_ENV;
+// const packageJson = require('./package.json');
+
+// const externals = Object.keys(packageJson.dependencies).reduce((acc, key) => {
+//   acc[key] = key;
+//   return acc;
+// }, {});
```
svalasovich, line 12: Do we need this?

packages/ddc-client/src/browser.ts
```diff
@@ -18,9 +18,10 @@ export {FileStorageConfig, KB, MB} from "@cere-ddc-sdk/file-storage";
 
 export class DdcClient extends CoreDdcClient {
 
-    static async buildAndConnect(options: ClientOptions, secretPhrase: string, encryptionSecretPhrase: string): Promise<DdcClient>;
-    static async buildAndConnect(options: ClientOptions, secretPhrase: string): Promise<DdcClient>;
-    static async buildAndConnect(options: ClientOptions, account: InjectedAccount, encryptionSecretPhrase: string): Promise<DdcClient>;
+    static async buildAndConnect(options: ClientOptions, accountOrSecretPhrase: string, encryptionSecretPhrase: string): Promise<DdcClient>;
+    static async buildAndConnect(options: ClientOptions, accountOrSecretPhrase: string): Promise<DdcClient>;
+    static async buildAndConnect(options: ClientOptions, accountOrSecretPhrase: InjectedAccount): Promise<DdcClient>;
+    static async buildAndConnect(options: ClientOptions, accountOrSecretPhrase: InjectedAccount, encryptionSecretPhrase: string): Promise<DdcClient>;
```
svalasovich, line 24: when type is just `string` -> parameter name secretPhrase, when `InjectedAccount` -> account. It's for better understanding parameters in IDE

svalasovich (approved):

packages/core/package.json
aelyseev: For the build, this doesn’t matter. The only difference is that, if a package is in the `dependencies` section it won't be included in the bundle, if it is in the `devDependencies` it will be included. 
Will skip for know, may be the point of the future improvements.

package.json
aelyseev: Fixed

packages/core/webpack.config.js
aelyseev: Nope, removed.

packages/ddc-client/src/browser.ts
aelyseev: Reverted to the prev version

PR #57, opened 2022-10-19
feat: Add integration tests

aelyseev:
- [x] remove session for the store requests


packages/proto/package.json
```diff
@@ -21,7 +21,7 @@
   "author": "",
   "license": "Apache-2.0",
   "scripts": {
-    "protoc": "npx protoc --ts_out ./src --proto_path ../../ddc-schemas/storage/protobuf $(find ../../ddc-schemas/storage/protobuf -name '*.proto' -type f | sed 's/.*\\///' | sort)",
+    "protoc": "protoc --ts_out ./src --proto_path ../../ddc-schemas/storage/protobuf $(find ../../ddc-schemas/storage/protobuf -name '*.proto' -type f | sed 's/.*\\///' | sort)",
```
svalasovich, line 24: Why do you remove npx? Will it work without installed protoc?

svalasovich (approved):

packages/proto/package.json
aelyseev: 1. New version 3 of `protoc` doesn’t require js or ts plugins
2. `npx protoc` won't work unless you have `protoc` installed

PR #59, opened 2022-10-24
feat: Expose scheme address

aelyseev:


packages/core/src/crypto/signature/Scheme.ts
```diff
@@ -7,10 +7,12 @@ import {SchemeInterface, SchemeName} from "./Scheme.interface";
 export class Scheme implements SchemeInterface {
     keyringPair: KeyringPair;
     name: SchemeName;
+    address: string;
 
     private constructor(keyringPair: KeyringPair, name: SchemeName) {
         this.keyringPair = keyringPair;
         this.name = name;
+        this.address = keyringPair.address;
```
svalasovich, line 15: Can we create `getter` than separate field?

PR #64, opened 2022-10-31
added trigger to start e2e tests

Isqander:


.github/workflows/release.yaml
```diff
@@ -34,3 +34,11 @@ jobs:
         run: npm run test
       - name: Publish
         run: node scripts/publish.js
+      - name: Trigger e2e tests
```
aelyseev, line 37: **issue:**
Wrong place for running e2e tests. First of all, these tests will run after publishing, i.e. errors could be published, secondly better to have these tests on ci, rather than on release. In this case, only proper code will be released.

aelyseev: Removed prev comment

aelyseev (approved):

.github/workflows/release.yaml
```diff
@@ -34,3 +34,11 @@ jobs:
...
+        run: |
+          curl \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.TRIGGER_GITHUB_TOKEN }}" \
+          https://api.github.com/repos/Cerebellum-Network/ddc-e2e-tests/dispatches \
+          -d '{"event_type":"sdk_release","client_payload":{"sdk":"golang"}}'
```
svalasovich, line 44: My example with docker image didn't work?

skambalin:
This PR looks outdated.

PR #65, opened 2022-11-02
feat: Add a retry mechanism for requests

aelyseev:
Add repeat attempts for `GET` requests


svalasovich (approved):
Looks good, but please think about GetFirstArgument using again

PR #66, opened 2022-11-04
feat: Verify piece data calculated cid with the cid in the request

aelyseev:


packages/content-addressable-storage/src/ContentAddressableStorage.ts
```diff
@@ -291,10 +292,22 @@ export class ContentAddressableStorage {
             );
         }
 
+        const pieceVerifyResult = await this.verifySignedPiece(pbSignedPiece, cid);
+        if (!pieceVerifyResult) {
+            throw new Error(
+                `Incorrect response. The cid [${cid}] provided in the request doesn't corresponds to expected`,
+            );
+        }
+
```
svalasovich, line 301: Let's add some configuration param, with default value always verify, but can be turned off.

svalasovich (approved):

packages/content-addressable-storage/src/ContentAddressableStorage.ts
aelyseev: Ok, will add

aelyseev: Discussed and decided to skip so far

PR #77, opened 2022-12-12
Extend SC with validation API

AndreiNavoichyk:


skambalin:
This PR looks outdated and the author is no longer available to tell if it is actual.

PR #79, opened 2023-01-10
feat: Add write attempts

aelyseev:


packages/file-storage/src/core/CoreFileStorage.ts
```diff
@@ -6,6 +6,7 @@ import {
     Tag,
     EncryptionOptions,
 } from '@cere-ddc-sdk/content-addressable-storage';
+import * as streamWeb from 'stream/web';
```
svalasovich, line 9: Will it work in browser and nodejs?

svalasovich (approved):

packages/file-storage/src/core/CoreFileStorage.ts
aelyseev: Fixed, thanks.

PR #86, opened 2023-03-13
Split publishing process

aelyseev:
### Content of the PR:

- [x] moved to `yarn` instead of `npm` as `yarn` works better with mono repositories.
- [x] removed sibling dependencies between packages, i.e. one can publish packages independently. To do this, one should change a version in the corresponding `package.json` file, and create a release as how was done in the past.
- [x] get rid of publishing of internal packages `core`, `proto`

I checked this PR manually, i.e. tested manually `@cere-ddc-sdk/content-addressable-storage` and `@cere-ddc-sdk/ddc-client`:
- [x] created instances of the content addressable storage, and the DDC client
- [x] creating a bucket with the DDC client
- [x] storing and reading of the arbitrary data

Also, RPC URLs for DEVNET and QANET have been changed.


skambalin:
The build and publish flow changed entirely using [Lerna.js](https://lerna.js.org/). This PR is outdated.

PR #92, opened 2023-06-27
Update SC addresses and metadata

skambalin:
- Updated SC addresses 
- Update SC metadata


packages/smart-contract/src/options/SmartContractOptions.ts
```diff
@@ -14,13 +14,13 @@ export const MAINNET: SmartContractOptions = {
...
     rpcUrl: 'wss://rpc.devnet.cere.network/ws',
-    contractAddress: '5DWvLrLFvFXTHETdMs3kRh9FqjhR6H6Hwt1QMnQviRLjKtms',
+    contractAddress: '6So8eqxMyWAxJ4ZZ2wCcJym7Cy6BYkc4V8GZZD9wgdCqWMQB',
     abi: ddcBucketAbi,
 };
 
 export const DEVNET: SmartContractOptions = {
     rpcUrl: 'wss://archive.devnet.cere.network/ws',
-    contractAddress: '6Ut3yTBNgmGBD9noJRaVHwUoFAswZ8SujnJiu6ChEs7Vi8ew',
+    contractAddress: '6SfBsKbfPUTN35GCcqAHSMY4MemedK2A73VeJ34Z2FV6PB4r',
```
yahortsaryk, line 23: I think we should use addresses [from this table](https://www.notion.so/cere/DDC-Infrastructure-e4dcbac6e3794d41878d80fbae08379b)

So, it should probably be:
for DEVNET 6SfBsKbfPUTN35GCcqAHSMY4MemedK2A73VeJ34Z2FV6PB4r
for TESTNET 6R2PF5gzKYbNkNLymTr8YNeQgWqNkE6azspwaMLZF2UHc1sg

yahortsaryk (changes requested):

packages/smart-contract/src/options/SmartContractOptions.ts
skambalin: Updated, thanks

yahortsaryk (approved):

PR #93, opened 2023-06-30
Implement missing SC API and cover with tests

skambalin:
- [x] Add blockchain node to the local test environment
- [x] Deploy DDC Bucket Contract to the node
- [x] Fill the contract with test data
- [x] Add missing contract methods
- [x] Cover all the methods with tests


ddc-test-cluster/docker-compose.yml
```diff
@@ -1,6 +1,15 @@
 version: "3"
 
 services:
```
evgeny-s, line 3: Can we switch to the redid image, which we built recently? Ask Pavel to help

ddc-test-cluster/redis/main.lua
```diff
@@ -46,8 +50,21 @@ local function update_node_aggregates(node_aggregation_redis_key, node_public_ke
             ['totalQueriesAcked'] = (acked and log_type == 3) and 1 or 0,
             ['totalBytesSent'] = bytes_sent,
```
evgeny-s, line 51: And in this case we should remove all the lua files

ddc-test-cluster/docker-compose.yml
skambalin: Good point. I will create another PR for this and merge it to `main` branch. Then pull the changes to this one.

skambalin: Update the test environment to use redis image

ddc-test-cluster/redis/main.lua
skambalin: Done

PR #96, opened 2023-07-11
feat: Update browser client

aelyseev:
Fixed export, which occurs error for a browser build.

Added manual version set up for `scripts/release.js`


packages/content-addressable-storage/package.json
```diff
@@ -1,64 +1,64 @@
 {
-    "name": "@cere-ddc-sdk/content-addressable-storage",
-    "description": "Content-addressable storage client",
-    "version": "1.7.6-rc.2",
-    "type": "module",
-    "repository": {
-        "type": "git",
-        "directory": "packages/content-addressable-storage",
-        "url": "git+https://github.com/Cerebellum-Network/cere-ddc-sdk-js.git"
+  "name": "@cere-ddc-sdk/content-addressable-storage",
```
skambalin, line 2: it is hard to tell what is changed in JSON files due to auto-format. it seems your IDE does not respect prettier configuration for JSON files. Can you please roll back the auto-formatted changes to follow prettier setup.

skambalin: Or maybe it is not related to IDE but the files were generated by script, so maybe we can adjust the script?

skambalin (approved):

PR #111, opened 2023-09-06
Get $CERE decimals value from blockchain properties

skambalin:
Use dynamic value of CERE token decimals from blockchain properties instead of hardcoding it to equal `10`

### DIFF
--- a/packages/smart-contract/src/SmartContract.ts
+++ b/packages/smart-contract/src/SmartContract.ts
@@ -29,8 +29,6 @@ import {
     NodeStatusInCluster,
 } from './types';
 
-const CERE = 10_000_000_000n;
-
 export class SmartContract extends SmartContractBase {
     private shouldDisconnectAPI = false;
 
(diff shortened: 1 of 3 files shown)


Argishtyfirst (approved):

packages/smart-contract/src/SmartContract.ts
```diff
@@ -80,8 +78,26 @@ export class SmartContract extends SmartContractBase {
         }
     }
 
+    toUnits(tokens: bigint) {
```
khssnv, line 81: The value is not expected to be changed frequently. It can be set during the initialization, on a first request to the RPC endpoint, or on the first access to the multiplier (using a Proxy object for instance). It would be much more efficient than this function. What do you think?

khssnv:
Are you sure it should go to the feature branch with a version, not to the main or another branch first?

packages/smart-contract/src/SmartContract.ts
skambalin:
```ts
api.registry.chainDecimals;
```
It is a static getter. I suppose the value is fetched from the chain on api connect. So in general I think the function is of a constant complexity since it does not do any async requests.

khssnv (approved):

skambalin:
@khssnv 
The contract class has changed a lot in the new version, so adding this changes to the old version will make 2 separate implementations. Given that we plan the migration in the upcoming days (e2e tests are already migrated afaik) it might be sub optional to support 2 different versions. Also it is not a critical bug, since on real envs the hardcoded value is correct.

PR #127, opened 2023-10-06
feat: hasPermission method

samvel33651:
hasPermission  method

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.9.1-rc.1",
+    "version": "1.9.1",
     "command": {
         "publish": {
             "directory": "{workspaceRoot}/{projectRoot}/package"
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-    "version": "1.9.0",
+    "version": "1.9.1-rc.1",
     "command": {
         "publish": {
             "directory": "{workspaceRoot}/{projectRoot}/package"
--- a/packages/smart-contract/src/SmartContract.ts
+++ b/packages/smart-contract/src/SmartContract.ts
@@ -318,6 +318,7 @@ export class SmartContract extends SmartContractBase {
     }
 
     async hasPermission(accountId: AccountId, permission: Permission) {
-        await this.submit(this.contract.tx.hasPermission, accountId, permission);
+        const result = await this.submit(this.contract.tx.hasPermission, accountId, permission);
+        return Boolean(result);
     }
 }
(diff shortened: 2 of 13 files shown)


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "1.9.0",
+    "version": "1.9.1-rc.1",
```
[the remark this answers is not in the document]
samvel33651: yeah I can use patch

packages/smart-contract/src/types/contractTypes.ts
```diff
@@ -31,6 +31,13 @@ export enum NodeStatusInCluster {
     OFFLINE = 'OFFLINE',
 }
 
+export enum PermissionType {
```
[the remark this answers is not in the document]
samvel33651: ok changing

tests/SmartContract.spec.ts
```diff
@@ -63,6 +63,15 @@ describe('Smart Contract', () => {
             await adminContract.revokeTrustedManagerPermission(user.address);
         });
 
+        test('has trusted manager permission', async () => {
+            await adminContract.grantTrustedManagerPermission(user.address);
+            const hasCMPermission = await adminContract.hasPermission(
+                user.address,
+                PermissionType.CLUSTER_MANAGER_TRUSTED_BY,
+            );
+            expect(hasCMPermission).toEqual('true');
```
skambalin, line 72: It also will be nice to convert 'true'/'false' to boolean inside the `hasPermission` method to make the result API more obviouse

skambalin: Just checked that the contract method `hasPermission` returns `boolean`, so it is not clear why you are checking it to equal `true` string?

Maybe:
```suggestion
            expect(hasCMPermission).toEqual(true);
```

skambalin (approved):

packages/smart-contract/src/SmartContract.ts
```diff
@@ -315,4 +316,9 @@ export class SmartContract extends SmartContractBase {
     async revokeTrustedManagerPermission(managerAddress: AccountId) {
         await this.submit(this.contract.tx.revokeTrustedManagerPermission, managerAddress);
     }
+
+    async hasPermission(accountId: AccountId, permission: Permission) {
+        const result = await this.submit(this.contract.tx.hasPermission, accountId, permission);
+        return Boolean(result);
```
skambalin, line 322: This conversion is not needed I suppose since the method should return `boolean` on its own, but it will not break anything anyway.

skambalin (approved):

idmitriev (approved):

PR #137, opened 2023-10-27
Storage Node high level API

skambalin:
High level API for storage node gRPC services. It adds new package `@cere-ddc-sdk/ddc` which can be used to communicate with DDC nodes (currently only storage nodes). Despite it is called hight level it is still for advanced use cases. Its API will be used in current `@cere-ddc-sdk/ddc-client` facade module as a replacement for the following packages:
- @cere-ddc-sdk/content-addressable-storage
- @cere-ddc-sdk/file-storage
- @cere-ddc-sdk/key-value-storage

P.S.
It is still work in progress and the API can change. The PR is an atomic change (better for code review) of a larger migration.

Not implemented yet features:
- Range access
- Proofs
- Browser support
- Lots of TODOs
- Some APIs are still not working (Large file upload, Range access)


Argishtyfirst (approved):

packages/ddc/src/DagApi/DagApi.ts
```diff
@@ -10,13 +10,13 @@ export class DagApi {
         this.api = new DagApiClient(transport);
     }
 
-    async store(request: PutRequest) {
+    async storeNode(request: PutRequest) {
```
upalinski, line 13: might be we can left it the same as in storage API? dagApi.put, dagApi.get?

during naming consideration in storage API i've references to IPFS naming: https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/DAG.md

packages/ddc/src/DagNode.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class Tag implements dag.Tag {
+    constructor(public key: string, public value: string) {}
+}
+
+export class DagNode implements dag.Node {
+    constructor(public bucketId: bigint, public data: Uint8Array, public links: Link[] = [], public tags: Tag[] = []) {}
+}
+
+export class DagNodeResponse extends DagNode {
```
upalinski, line 15: why responses extends models and includes all fields + cid?

packages/ddc/src/FileApi/FileApi.ts
```diff
@@ -64,7 +64,7 @@ export class FileApi {
         return cid;
     }
 
-    read(request: GetFileRequest) {
+    readPiece(request: GetFileRequest) {
```
upalinski, line 67: getField? Or readFile?

in some places you did renaming, but models are used from grpc package. Because of it your methods looks a little bit inconsistent. e.g. here method named `readPiece` but accepts `GetFileRequest`.

packages/ddc/src/StorageNode.ts
```diff
@@ -0,0 +1,110 @@
...
+import {DagNode, DagNodeResponse} from './DagNode';
+
+type NamingOptions = {
+    name?: string;
+};
+
+type PieceStoreOptions = NamingOptions & {};
+type DagNodeStoreOptions = NamingOptions & {};
+
+export class StorageNode {
```
upalinski, line 16: what is purpose of this component? Looks like it's neither low level nor high level.

upalinski: in case the idea was to provide high level API, then i think there shouldn't be methods like storeRawPiece, storeMultipartPiece. It should be storeFile that accepts file or stream and readFile that returns the stream as well. Otherwise it exposes internals

packages/ddc/src/DagApi/DagApi.ts
skambalin: Updated the naming to align it with gRPC names

packages/ddc/src/FileApi/FileApi.ts
skambalin: It actually looked a bit confusing to me, why we have `putPiece` names but getter is called `getFile`. In my understanding, `File` is a more high level thing. For example it might have some stats or metadata. But if we work with simple data it can be called piece for `put` and `read` operations.

But anyway, I updated the naming for this low level API to align it with gRPC names.

packages/ddc/src/StorageNode.ts
skambalin: The purpose of this class is to configure all 3 APIs (create a single transport for all) and provide flat API for all 3 services. In the current version it mostly mirroring the low level API, but later it will be extended. eg. we will need to authenticate the grpc transport.

And in general this class is facade for the services API. I suppose the services API will rarely use directly.

skambalin: I would not like to overload this class API with file related logic. For this we can use another package `@cere-ddc-sdk/file-storage` (or we can rename it to a shorter `@cere-ddc-sdk/fs`).

In my opinion, working with files is a relatively complex logic with should go in a separate package. Apart from simple data pieces, files are more complex, they should have some associate metadata and even can have extended storing algorithm (just a thought for now): 
- Revisions
- Metadata (size, mimetype, etc...)
- Tags

The `fs` package can also provide several extra ways of storing files, eg.:
- From local FS (NodeJs)
- From file reader (Browser)

packages/ddc/src/FileApi/FileApi.ts
upalinski: it's called FileApi and file can be represented as either raw piece (for small files) or as a multipart piece (for large files). This difference is important when you upload a file because the flow of storing is different. But after uploading you can treat it as a single file and access it by the single identifier. 

it's how i'm thinking about it... might be my vision isn't correct. Or might be later we will add a methods to access raw piece and multipiece directly... 

btw, thank you for the alignment

packages/ddc/src/StorageNode.ts
upalinski: okey, thank you. got it.

upalinski (approved):

PR #138, opened 2023-10-30
Add blockchain package

idmitriev:
- Add new `blockchain` package with pallet APIs
 - Add tests for `blockchain` package
 - Use `blockchain` package for routing in DDC client


packages/blockchain/src/DDCClustersPallet.ts
```diff
@@ -0,0 +1,75 @@
...
+    removeCdnNodeFromCluster(clusterId: ClusterId, cdnNodePublicKey: CdnNodePublicKey) {
+        return this.apiPromise.tx.ddcClusters.removeNode(clusterId, {CDNPubKey: cdnNodePublicKey}) as Sendable;
+    }
+
+    removeStorageNodeFromCluster(clusterId: ClusterId, storageNodePublicKey: StorageNodePublicKey) {
+        return this.apiPromise.tx.ddcClusters.removeNode(clusterId, {StoragePubKey: storageNodePublicKey}) as Sendable;
+    }
+}
+
+export type ClusterId = HexString;
```
skambalin, line 66: What do you think if we move all these domain types to a single file `types.ts` - to be able to see the whole picture of domain types (Cluster, BucketId, Account, etc...).

packages/ddc/src/Router/Router.ts
```diff
@@ -26,16 +24,37 @@ export class Router {
...
+            if (cluster == null) {
+                throw new Error(`Failed to get info for cluster ${bucket.clusterId} on blockchain`);
+            }
+            const nodeKeys = await this.config.blockchain.ddcClusters.listNodeKeys(cluster.clusterId);
+            const nodeKey = nodeKeys[Math.floor(Math.random() * nodeKeys.length)];
+            const node = await this.config.blockchain.ddcNodes.findStorageNodeByPublicKey(nodeKey);
+            if (node == null) {
+                throw new Error(`Failed to get info for node ${nodeKey} on blockchain`);
+            }
+            return new StorageNode({rpcHost: node.params.rpcHost, signer: this.config.signer});
```
skambalin, line 48: I suppose we can cache something from this block, eg by `bucketId`. In most use cases the SDK will work with a single bucket or cluster, imo. So fetching all this data each time a storage node is requested might be an overhead. We can do it later though.

packages/blockchain/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "description": "Cere Blockchain client",
+  "type": "module",
+  "main": "./dist/index.cjs",
+  "types": "./dist/types",
+  "repository": {
+    "type": "git",
+    "directory": "packages/blockchain",
+    "url": "git+https://github.com/Cerebellum-Network/cere-ddc-sdk-js.git"
+  },
+  "scripts": {
```
skambalin, line 14: You need to add `export` field here to support `module` imports. Check some other packages for the example of what to add. eg `ddc-client`

packages/blockchain/src/string-utils.ts
```diff
@@ -0,0 +1,22 @@
+import {HexString} from './Blockchain';
+
+export function hexEncode(s: string) {
```
skambalin, line 3: I suppose there should be ready functions in `@polkadot/util` for this

skambalin (approved):

packages/blockchain/src/DDCClustersPallet.ts
idmitriev: Sure, sounds right

packages/ddc/src/Router/Router.ts
idmitriev: Yeah, caching buckets+clusters seems logical

packages/blockchain/package.json
idmitriev: thanks!

packages/blockchain/src/string-utils.ts
idmitriev: yep

PR #141, opened 2023-11-01
File storage package upgrade

skambalin:
- Update the `@cere-ddc-sdk/file-storage` to use new architecture
- Add tests for small and large files


Argishtyfirst (approved):

packages/ddc/src/Cid.ts
```diff
@@ -8,10 +8,10 @@ export class Cid {
     }
 
     toString() {
-        return typeof this.cid === 'string' ? this.cid : base32.stringify(this.cid);
+        return typeof this.cid === 'string' ? this.cid : base32.stringify(this.cid, {pad: false});
```
upalinski, line 11: There is a better future-proofing approach to encode CIDs. 

We took a CID idea from Protocol Labs and CID's in IPFS. The idea is that it should be self descriptive values that can evolve over the time. The encoding type (base32, base58 etc.) should be encoded in CID itself. So later you can change encoding and understand the encoding type by start parsing CID.

Please, take a look here: https://www.notion.so/ADR-Protocol-interoperability-and-future-proofing-aaccc53547e84b5e93a9ca334d68e372

Other links:
https://github.com/multiformats/cid
https://github.com/multiformats/multibase

packages/ddc/src/streams/splitStream.ts
```diff
@@ -0,0 +1,38 @@
...
+    stream: ReadableStream<Uint8Array>,
+    chunkSize: bigint,
+    mapper: SplitStreamMapper<T>,
+) => {
+    const out: any = [];
+
+    let correntSize = 0n;
+    let currentWriter: Writer | undefined;
+
+    for await (const chunk of stream) {
```
upalinski, line 15: do i correctly understand that we have 2 concepts of chunks here? One is part of stream send by client to SDK and other is a chunk to be send in grpc stream to server?

packages/ddc/src/streams/splitStream.ts
```diff
@@ -0,0 +1,38 @@
...
+            currentWriter = writable.getWriter();
+            correntSize = 0n;
+
+            out.push(mapper(readable, out.length));
+        }
+
+        await currentWriter.write(chunk);
+        correntSize += BigInt(chunk.byteLength);
+
+        if (correntSize === chunkSize) {
```
upalinski, line 28: are we sure that cUrrentSize is always aligned to chunkSize? Can't there be a situation when currentSize > chunkSize? 

Shouldn't we check size and in case it's more than currentSize split it to fill full chunkSize in currentWriter and initalise new one and put part of the chunk that is left?

packages/file-storage/src/FileStorage.ts
```diff
@@ -0,0 +1,69 @@
...
+export type FileStoreOptions = PieceStoreOptions;
+
+export class FileStorage {
+    private storageNode: StorageNode;
+
+    constructor({storageNode}: FileStorageConfig) {
+        this.storageNode = storageNode;
+    }
+
+    private async storeLarge(bucketId: bigint, file: File, options?: FileStoreOptions) {
```
upalinski, line 28: do we really want to expose different methods for different file sizes? As for me there should be single method and details hidden inside SDK.

Otherwise clients needs to distinguish what is large file and what is small file

upalinski: oh sorry. Didn't notice that it's private methods :) usually public methods are at the top of the file :) everything is fine. Only might be naming can be aligned to the storage concepts? storeLarge -> storeMultipartPiece, storeSmall -> storeRawPiece?

But renaming is up to you. Only if you also see an advantage, otherwise feel free to left it as is.

packages/file-storage/src/FileStorage.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private async storeSmall(bucketId: bigint, file: File, options?: FileStoreOptions) {
+        const piece = new Piece(file.content);
+
+        return this.storageNode.storePiece(bucketId, piece, options);
+    }
+
+    async store(bucketId: bigint, file: File, options?: FileStoreOptions) {
+        if (file.size && file.size > MAX_PIECE_SIZE) {
```
upalinski, line 57: looks like file.size is optional here. It means that if client didn't specify it and tries to upload large file he would get an error after trying an upload... might be make sense to make it required?

packages/ddc/src/Cid.ts
skambalin: Added a [related task](https://www.notion.so/cere/Properly-generate-CIDs-2ebb2b046b45482da61dab79b9c68339?pvs=4)

packages/ddc/src/streams/splitStream.ts
upalinski: also in tests you produce clients stream that is 64 KB and then pieceSize is equally dividable by it. But i think real client stream can by fully randomised like send 5 KB, 3 KB, 100 KB, 8.4 KB etc. Right?

packages/ddc/src/streams/splitStream.ts
skambalin: Currently it is the same stream of data which later goes to gRPC. So if the client provide a stream with chunks of 64kb (eg. `fs.createReadableStream()`) these chunks will be sent to gRPC. We can buffer chunks to make them bigger before sending  to gRPC (or smaller) later as a form of optimization, if there is no strict requirements for the chunk size currently which can be a blocker.

packages/file-storage/src/FileStorage.ts
skambalin: The method is `private` the same as for `storeSmall` - not exposed. These both methods called from public `store` method

packages/file-storage/src/FileStorage.ts
skambalin: I tested a MultipartPiece with a single chunk (which is the case when user uploads small file without `size`) and it works well. So my assumption was that it just not optimal way of storing small files, but working way. So I made `size` optional. Is storing a multipart pieces with a single piece a really bad practice and we should the size required to mitigate it, or it is just an storage optimization thing and we can keep it optional?

packages/ddc/src/streams/splitStream.ts
skambalin: Yes. Good point. I suppose most of the time it will be 64 KB or 16 KB - because it is default values for standard file read streams, but we need to cover variable chunks as well. Will fix.

packages/file-storage/src/FileStorage.ts
upalinski: i would say it's a bad practice for storage. Might be latter we will not allow it to do (just don't know the proper way to control it at the moment).

Also according to your if statement:   if (file.size && file.size > MAX_PIECE_SIZE) {

looks like if file size isn't specified - it will upload it through storeSmallFile. No? otherwise you need to put `||` instead of `&&`. Might be i'm wrong

packages/ddc/src/streams/splitStream.ts
skambalin: Fixed

packages/file-storage/src/FileStorage.ts
skambalin: Will do it `required` then.

skambalin: Done

upalinski (approved):

PR #142, opened 2023-11-02
Migrate ddc-client package to new storage architecture

skambalin:
Migrate `ddc-client` package to use new storage (MVP version) .

The PR also includes changes in other pacakges:
- file-storage: Transform input stream to fixed 64 kb chunk stream
- ddc - replace `bigint` with `number`


Argishtyfirst (approved):

packages/ddc-client/src/DdcClient.ts
```diff
@@ -26,65 +16,34 @@ import {
...
-    ): Promise<DdcClient> {
-        encryptionSecretPhrase = encryptionSecretPhrase != null ? encryptionSecretPhrase : secretPhrase;
-        const clientCreateOptions = initDefaultClientOptions(options);
-
-        const scheme = isSchemeName(clientCreateOptions.scheme)
-            ? await Scheme.createScheme(clientCreateOptions.scheme, secretPhrase)
-            : clientCreateOptions.scheme;
+        readonly smartContract: SmartContract,
+        private scheme: Scheme,
+        private fileStorage: FileStorage,
```
upalinski, line 30: might be we should hide from out client initialisation of low-level APIs?

packages/ddc-client/src/DdcClient.ts
```diff
@@ -26,65 +16,34 @@ import {
...
+        private router: Router,
+    ) {}
 
-        const smartContract = await SmartContract.buildAndConnect(secretPhrase, options.smartContract);
-        const caStorage = await ContentAddressableStorage.build(
-            {...clientCreateOptions, smartContract, scheme},
-            secretPhrase,
-        );
+    static async buildAndConnect(config: DdcClientConfig, secretPhrase: string) {
+        const scheme = await Scheme.createScheme('sr25519', secretPhrase);
```
upalinski, line 35: before the scheme (sr25519) was it options. why do we removed it and decided to use hard code?

packages/ddc-client/src/DdcUri.ts
```diff
@@ -0,0 +1,8 @@
+export type DdcEntity = 'file' | 'dag-node';
+
+export type FileUri = DdcUri<'file'>;
+export type DagNodeUri = DdcUri<'dag-node'>;
+
+export class DdcUri<T extends DdcEntity = DdcEntity> {
+    constructor(readonly bucketId: number, readonly cid: string, readonly entity: T) {}
```
upalinski, line 7: actually, content type can be extracted from CID. no need to sepcify separate parameter. please, check this doc https://www.notion.so/cere/ADR-Protocol-interoperability-and-future-proofing-aaccc53547e84b5e93a9ca334d68e372

and out encoding component https://github.com/Cerebellum-Network/ddc-storage-node/blob/8322a64ece2fd11febe83fddcb7c337aeb05bede/pkg/cid/cid.go#L27C1-L27C1

packages/ddc/src/Router/Router.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export type RouterNode = {
+    rpcHost: string;
+};
+
+export class Router {
+    constructor(private nodes: RouterNode[]) {}
+
+    async getNode(operation: RouterOperation) {
```
upalinski, line 23: looks like you don't use RouterOperation here. What is a purpose of this enum?

skambalin: For now it is a very temporary approach just to be able to integrate into real env without blockchain pallets. In general, based on the operation later will decide what node to choose (as it was done in our DAC v3 attempt) in future. But we can postpone the discussion, I think.

upalinski: not sure it make sense in our new approach  (v4 i guess ? :) ) okey, we can return to this discussion when we will implement real rotuing via pallet

packages/ddc-client/src/DdcUri.ts
skambalin: It is also needed for build time type checking and type hinting (typescript).

skambalin: But it is nice to know in general, that we can detect this from cid

packages/ddc-client/src/DdcClient.ts
skambalin: The constructor is protected, so I don't think we expose the API here. And also I am not sure that `FileStorage` is a low level API. The package can be used directly if the client don't need bucket/customer management and only wants to read/store files. At least it how it was in past and I think it is a nice option.

packages/ddc/src/Router/Router.ts
skambalin: One example might be is read and store. When we want read data from CDN and store to Storage nodes. Router can decide what node to provide for the operation.

PR #144, opened 2023-11-03
Support latest storage node API types changes

skambalin:
These API changes suported
- CIDs are always of `bytes` type (`Uint8Array`)
- Bucket IDs are always of `uint64` type (`number`)


Argishtyfirst (approved):
lgtm

PR #149, opened 2023-11-08
GRPC web over envoy proxy PoC

skambalin:
GRPC native transport cannot be use in browser due to browsers limitation of HTTP2 API. There is alternative protocol called [GRPC Web](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) which aims to bring GRPC to web. This PoC is an attempt to switch the SDK to use GRPC Web as a transport level in browsers.

The SDK is configured to use emulated browser `fetch` API in JEST environment, so if it works in this env it is high probably will work in real browser.

What included:
- Temp [GRPC web transfer](https://github.com/timostamm/protobuf-ts/blob/main/MANUAL.md#grpc-web-transport) setup always pointing to one storage node (`ddc-storage-node-1`)
- [Envoy proxy configuration](https://grpc.io/docs/platforms/web/basics/) for `ddc-storage-node-1`. It proxies GRPC Web requests to GRPC Native server on the storage node


skambalin:
As it can be seen from [the test results](https://github.com/Cerebellum-Network/cere-ddc-sdk-js/actions/runs/6795186124/job/18472826103) the current GRPC Web tooling does not support `client streams` which leads to the following results by APIs:
- Dag API - functional
- CNS API - functional
- File API - non functional

File API uses client streaming to upload raw pieces and after the last changes `getFile` also uses client streams (duplex actually). So it makes the entire API non functional in browsers using ready-to-use toolset.

skambalin:
I found mentions of several attempts of making client streaming work, but could not found a working solution so far. [here](https://grpc.io/blog/state-of-grpc-web/) is the state of GRPC in browsers page where we have links to various solutions. Plus [here](https://github.com/johanbrandhorst/grpc-web-compatibility-test) we have a project where some of the solution tested.

Both links are quite old, but it is a good starting point from where to start looking for a solution.

skambalin:
There are several attempts to make client stream working, using various browser APIs, (eg. fetch streams, WebTransport, WebSockets), but all of them have limitations and developed by different teams. So a more deep research needed here.

Examples:
- https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/docs/transport.md
- https://github.com/grpc/grpc-web/blob/master/doc/streaming-roadmap.md

skambalin:
Since the native GRPC transport cannot be used in browser, in any case we need support of GRPC Web on our nodes. I currently use [envoy proxy](https://www.envoyproxy.io/) for this PoC but there are several GoLang options I found:
- https://github.com/improbable-eng/grpc-web/tree/master/go/grpcweb
- https://github.com/improbable-eng/grpc-web/tree/master/go/grpcwebproxy

skambalin:
DDC SDK Playground app demo recording:
https://github.com/Cerebellum-Network/cere-ddc-sdk-js/assets/56070785/7f4ffaf6-d7cd-4ad1-b0d6-6acd1b2af4c7

skambalin:
As it can be seen on the demo recording, the PoC proofed that it is possible to upload file using gRPC Web protocol over WebSocket transport.

In the demo 150 MB file has been uploaded for 15 seconds and downloaded for 6.3 seconds.

skambalin:
During the real file upload in the demo, CNS API is used under the hood to store the file name.

skambalin:
How to run the playground:
```sh
git clone git@github.com:Cerebellum-Network/cere-ddc-sdk-js.git
cd ./cere-ddc-sdk-js/
npm install
npm run compile
npm run build
npm run playground
```

upalinski (approved):

PR #158, opened 2023-11-17
Refactor packages dependencies and add CI tests

skambalin:
- Moved `Signer` implementation to `blockchain` package to keep all blockchain related  deps in one place
- Added `CI` test project to run tests on built version of packages
- Added `Dev` test project to run tests on source code
- Added config presets support to `file-storage` package to align it with `ddc-client` package

### DIFF
--- a/packages/ddc-client/src/DdcUri.ts
+++ b/packages/ddc-client/src/DdcUri.ts
@@ -8,7 +8,7 @@ export type DdcUriOptions = {
 };
 
 export class DdcUri<T extends DdcEntity = DdcEntity> {
-    readonly cid?: string;
+    readonly cid: string = '';
     readonly name?: string;
 
     constructor(readonly bucketId: BucketId, cidOrName: string, readonly entity: T, options?: DdcUriOptions) {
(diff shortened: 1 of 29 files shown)


packages/ddc-client/src/DdcUri.ts
```diff
@@ -8,7 +8,7 @@ export type DdcUriOptions = {
 };
 
 export class DdcUri<T extends DdcEntity = DdcEntity> {
-    readonly cid?: string;
+    readonly cid: string = '';
```
idmitriev, line 11: Why not undefined as a default value?

skambalin: It is a quick fix for the report from S&T. The optional type of the field makes it inconvenient to get CID right after storing a file. In most cases this property will have CID value. The only case when it will be empty is when you want to access a file or DagNode by its name:
```ts
await client.read(new FileUri(bucketId, 'somefileName.txt'));
```

But I suppose it does not make much sense to try getting CID in this particular case.

So I decided to add this quick fix, instead of more complex types hierarchy which requires adding a new type `StoredDdcUri` and its variants.

PR #159, opened 2023-11-17
Update documentation

skambalin:
Readme files:

- [x] Root
- [x] DDC Client
- [x] File Storage
- [x] DDC
- [x] Blockchain


skambalin:
Added stub documentation for utility packages for now, will granularly update them in future.

packages/blockchain/README.md
```diff
@@ -1,5 +1,7 @@
 # @cere-ddc-sdk/blockchain
 
+> The package is used internally in [@cere-ddc-sdk/ddc-client](/packages/ddc-client)
```
upalinski, line 3: client's will need this package to create bucket, top up account and so on. Right?

I think this sentence can confuse developers and they can think that is's internal-only package. 

Just shared thoughts, feel free to left as is

skambalin: @upalinski 
`ddc-client` will expose all these methods (forwarding them from `blockchain` package) when they are available and the `blockchain` package itself will contain more API then needed for ordinary users. The `blockchain` package will be used directly by our internal tools, eg. Cluster Manager.

Anyway we will extend the documentation after the initial release. Right now it is a stab README just in case someone will find this package on NPM.

upalinski (approved):

PR #162, opened 2023-11-21
Align bucketId type

skambalin:
- Export blockchain types (`BucketId`, `ClusterId`, ...) from. `ddc-client` package for easier access
- Align all bucket id usages to blockchain `BucketId` type (JS `bigint`)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -52,6 +52,9 @@ jobs:
             - name: Build packages
               run: npm run build
 
+            - name: Check types
+              run: npm run check-types:ci
+
             - name: Run integration tests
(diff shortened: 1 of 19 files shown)


skambalin:
@upalinski 
I had to patch `proto` schemes again to make `bucketId` of `bigint` type and the rest of `uint64` of type `number`. It is an intentional change. When we move the `proto` files to some common place we will need to decide if it is ok to add some JS specific directive there or I will always need to patch them on the SDK side.

idmitriev (approved):

upalinski:
@skambalin i think it's ok to put JS specific tags to proto spec once it's in some common repo. Sorry for delaying it, i remember about the issue. We'll do it

PR #177, opened 2023-11-29
Playground improvements and fixes

skambalin:
- Added links to file downloads
- Add support for long file names (DAG API + CNS API)
- Fixed several issues

### DIFF
--- a/playground/src/Playground.tsx
+++ b/playground/src/Playground.tsx
@@ -1,7 +1,7 @@
 import { useCallback, useState } from 'react';
 import { useDropzone } from 'react-dropzone';
 import { LoadingButton } from '@mui/lab';
-import TaskIcon from '@mui/icons-material/Task';
+import FileIcon from '@mui/icons-material/InsertDriveFileOutlined';
 import {
   File,
   Signer,
@@ -82,6 +82,10 @@ export const Playground = () => {
   const getFileUrlByName = (name: string) => [bcPresets[selectedBc].cdn, bucketId, cnsName, name].join('/');
   const getFileUrlByCid = (cid: string) => [bcPresets[selectedBc].cdn, bucketId, cid].join('/');
   const isCompleted = !!realFileCid && !!randomFileCid;
+
+  const handleSkip = useCallback(() => {
+    setStep(step + 1);
+  }, [step]);
 
   const handleConnectWallet = useCallback(async () => {
     setStep(1);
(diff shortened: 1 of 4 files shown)


playground/src/Playground.tsx
```diff
@@ -37,16 +51,18 @@ const Dropzone = styled(Box)(({ theme }) => ({
 }));
 
 const bcPresets = {
-  devnet: DEVNET,
-  testnet: TESTNET,
-  mainnet: MAINNET,
+  devnet: { ...DEVNET, cdn: 'https://ddc-devnet.cloud' },
```
upalinski, line 54: why CDN?

playground/src/Playground.tsx
```diff
@@ -105,14 +123,13 @@ export const Playground = () => {
...
     try {
-      const uri = await client!.store(bucketId, file, {
-        name: acceptedFile.name,
-      });
+      const file = new File(acceptedFile.stream(), { size: acceptedFile.size });
+      const uri = await client!.store(bucketId, file);
+      const fileLink = new Link(uri.cid, acceptedFile.size, acceptedFile.name);
+
+      const dagNode = new DagNode(JSON.stringify({ updateTime: Date.now() }), [fileLink]);
+      await client!.store(bucketId, dagNode, { name: cnsName });
```
upalinski, line 132: why do you need DAG here?

playground/src/Playground.tsx
skambalin: @upalinski 
The property is only for the playground app and is not exposed to SDK API. So technically can be anything. What name you suggest? How this entry point is called on the architectural level? `nodeProvider` or `entryPoint`... I can call it `ddcBaseUrl`.

playground/src/Playground.tsx
skambalin: To be able to upload a real file with long name and then access it by the name using HTTP

upalinski: looks like a hack. Because at first, our protocol doesn't support long name, and you can provide optional field in UI (e.g. file name) and either provide a link by CID or by short name specified. At second, you use static cnsName, that you attach to DAG node to which you attach file. It means that you always overwrite CNS name, and users (our internal) will loose their fiels because links will become invalid after someone's else upload.

playground/src/Playground.tsx
upalinski: don't know how to name it better. baseUrl is fine. Later we can think about naming. Might be just url. Or gateway url. Will see.

playground/src/Playground.tsx
skambalin: Later I add a `Link` to the `DagNode` which assigned to the constant `cnsName`. So the long name of the uploaded file foes to `Link(cid, size, name)`. It is how Gaming upload their game folders.

PR #179, opened 2023-12-05
Implement on-chain topology routing

skambalin:
- Use on-chain topology to select nodes for NodeJs and browser (with ssl)
- Refactor Router logic to split static and blockchain strategies


packages/ddc/src/Router/BlockchainStrategy.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  private async getClusterNodes(clusterId: ClusterId) {
+    if (this.clusterNodes.has(clusterId)) {
+      return this.clusterNodes.get(clusterId)!;
+    }
+
+    /**
+     * TODO: Discuss with the team if it is possible to get nodes with props from the blockchain by clusterId
+     */
+    const allNodes = await this.blockchain.ddcNodes.listStorageNodes();
```
idmitriev, line 57: `DDCClustersPallet` has `listNodeKeys(clusterId: ClusterId)`
but then you have to do multiple `find{Storage/Cdn}NodeByPublicKey` queries
still might be more efficient, than listing all the nodes 
cc @yahortsaryk

skambalin: @idmitriev unfortunately it will not be just several calls, in theory it can be `N` calls since I need to find a node with `ssl` support and it can be only one of `N`. So we need to extend the palette API, but in my opinion it is scalability question which is not currently a blocker, given we have eg. up to 1000 nodes.

idmitriev (approved):

PR #183, opened 2023-12-08
Account balance

idmitriev:
Cluster management needs this API

### DIFF
--- a/packages/blockchain/src/Blockchain.ts
+++ b/packages/blockchain/src/Blockchain.ts
@@ -8,6 +8,7 @@ import { DDCStakingPallet } from './DDCStakingPallet';
 import { DDCCustomersPallet } from './DDCCustomersPallet';
 import { AccountId } from '@polkadot/types/interfaces/runtime';
 import { formatBalance } from '@polkadot/util';
+import { AccountInfo } from '@polkadot/types/interfaces';
 
 export type BlockchainConnectOptions = {
   account: IKeyringPair;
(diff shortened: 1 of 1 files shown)


packages/blockchain/src/Blockchain.ts
```diff
@@ -89,6 +91,29 @@ export class Blockchain {
...
     return new Blockchain(api, options.account);
   }
+
+  formatBalance(balance: string | number | bigint) {
+    const chainDecimals = this.apiPromise.registry.chainDecimals[0];
+    formatBalance.setDefaults({ unit: 'CERE' });
+    return formatBalance(balance, { withSiFull: true, decimals: chainDecimals });
+  }
+
+  async getAccountFreeBalance(accountId: AccountId) {
```
skambalin, line 101: I would suggest a shorter version 
```ts
async getAccountFreeBalance(accountId: AccountId) {
    const { data } = await this.apiPromise.query.system.account<AccountInfo>(accountId);

    return data.free.toBigInt();
}
```
where `AccountInfo` comes from here
```ts
import { AccountInfo } from '@polkadot/types/interfaces';
```

packages/blockchain/src/Blockchain.ts
```diff
@@ -89,6 +91,29 @@ export class Blockchain {
 
     return new Blockchain(api, options.account);
   }
+
+  formatBalance(balance: string | number | bigint) {
+    const chainDecimals = this.apiPromise.registry.chainDecimals[0];
+    formatBalance.setDefaults({ unit: 'CERE' });
+    return formatBalance(balance, { withSiFull: true, decimals: chainDecimals });
```
skambalin, line 98: Can it work without setting settings globally? The SDK can be used in a project with different defaults, but the project and SDK can use the same deduped `@polkadot/util` package

```ts
formatBalance(balance: string | number | bigint) {
    const [chainDecimals] = this.apiPromise.registry.chainDecimals;

    return formatBalance(balance, { withSiFull: true, decimals: chainDecimals, withUnit: 'CERE' });
  }
```

PR #188, opened 2023-12-13
Blockchain writes signer

idmitriev:
- allow blockchain reads without specifying account
 - provide (optional) signer for blockchain writes

### DIFF
--- a/packages/blockchain/src/Blockchain.ts
+++ b/packages/blockchain/src/Blockchain.ts
@@ -1,6 +1,5 @@
 import { ApiPromise, WsProvider } from '@polkadot/api';
-import { SignerOptions } from '@polkadot/api/types';
-import { IKeyringPair } from '@polkadot/types/types';
+import { AddressOrPair, SignerOptions } from '@polkadot/api/types';
 import { Index } from '@polkadot/types/interfaces';
 import { SubmittableExtrinsic } from '@polkadot/api-base/types';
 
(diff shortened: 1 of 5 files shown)


packages/blockchain/src/Blockchain.ts
```diff
@@ -9,30 +9,24 @@ import { DDCClustersPallet } from './DDCClustersPallet';
 import { DDCStakingPallet } from './DDCStakingPallet';
 import { DDCCustomersPallet } from './DDCCustomersPallet';
 
-export type SendOptions = Pick<Partial<SignerOptions>, 'nonce'> & {
-  account?: IKeyringPair;
+export type SendOptions = Pick<Partial<SignerOptions>, 'nonce' | 'signer'> & {
+  account: IKeyringPair | string;
```
skambalin, line 13: Probably we can use `AddressOrPair` type instead

```ts
import { AddressOrPair } from '@polkadot/api/types';
```

packages/blockchain/src/Blockchain.ts
```diff
@@ -43,16 +37,16 @@ export class Blockchain {
     return !!this.apiPromise.isReady;
   }
 
-  async getNextNonce() {
-    const nonce = await this.apiPromise.rpc.system.accountNextIndex<Index>(this.account.address);
+  async getNextNonce(address: string) {
```
skambalin, line 40: To make the input params closer to `accountNextIndex`

```suggestion
  async getNextNonce(address: string | AccountId) {
```

skambalin (approved):

PR #191, opened 2023-12-16
Fix/ddc staking method naming

idmitriev:


packages/blockchain/src/DDCClustersPallet.ts
```diff
@@ -19,7 +19,7 @@ export class DDCClustersPallet {
     return entries.map(([, clusterOption]) => clusterOption.toJSON() as unknown as Cluster);
   }
 
-  async listNodeKeys(clusterId: ClusterId) {
+  async filterNodeKeysByClusterId(clusterId: ClusterId) {
```
skambalin, line 22: A braking change... Don't we use this method already in e2e or somewhere else? After stable release we will need to use deprecation strategy for such refactoring. For now we just need to don't forget to fix it in other projects.

skambalin (approved):

packages/blockchain/src/DDCClustersPallet.ts
idmitriev: Not used in e2e, will check other repos

PR #195, opened 2023-12-20
feat: Select node for operation based on nodes mode priority

skambalin:
- Update router to use node mode priority table during node selection for an operation
- Update local test environment to include all types of nodes
- Add router tests


packages/ddc/src/Router/PriorityStrategy.ts
```diff
@@ -0,0 +1,52 @@
...
+const priorityMap: OperationPriorityMap = {
+  [Operation.READ_DAG_NODE]: {
+    [Mode.Full]: 1,
+    [Mode.Cache]: 1,
+    [Mode.Storage]: 2,
+  },
+
+  [Operation.STORE_DAG_NODE]: {
+    [Mode.Full]: 1,
+    [Mode.Storage]: 1,
```
upalinski, line 18: lets priorities Full over Storage for store operation.

For now it will be like a hack, to sent less request to nodes located in Contabo cloud.

upalinski (approved):

packages/ddc/src/Router/PriorityStrategy.ts
skambalin: In this case Contabo will never be used by SDK for store operations. Is it ok?

skambalin: Done

PR #206, opened 2024-01-08
feat(ddc): Add DAG Node content hash validation

skambalin:
- Add DAG Node hash validation
- Update tests to cover the feature


packages/ddc/src/FileApi/FileApi.ts
```diff
@@ -50,6 +50,7 @@ export class FileApi {
     this.options = {
       ...options,
       enableAcks: options.enableAcks ?? !!options.signer, // ACKs are enabled by default if signer is provided
+      authenticate: options.authenticate ?? false, // TODO: Implement proof validation and enable authentication by default
```
upalinski, line 53: it's better to keep authentication disable by default because it produces additional load on the network and not everyone will care about authentication (also it's not needed in case of trusted providers) so it's better to left this options only for people who knows what they want (advanced clients)

skambalin: Made it `false` by default

PR #209, opened 2024-01-08
feat(blockchain): Add method to get current block number

idmitriev:
Required for cluster management


skambalin:
@idmitriev you need to sync with the target branch to fix tests. Also could you update the PR title to make it more descriptive - it will go to chagelog automatically. Maybe mention that you added a method to get current block number...

skambalin (approved):

PR #223, opened 2024-01-29
fix(ddc): Add gRPC request timeout to fallback from frozen nodes

skambalin:
- Add grpc requests inactivity timeout (30 seconds by default) 
- Add `DEADLINE_EXCEEDED` and `CANCELED` to re-try codes


skambalin:
It appeared then `grpc` timeout option is applied to requests from start to status response. Which does not work for us since the upload/download process can take long. We need to figure out another way to detect nodes freezes

skambalin:
Added custom client-side abortion  controller which tracks storage node activity (eg. new messages in a stream) and if the node does not send anything for 30 sec (by default) - the operation is rejected and re-tried with another node.

upalinski (approved):

PR #225, opened 2024-02-01
feat(blockchain): Add change bucket parameters method

aleksey-svetliskiy-cere:


skambalin:
Can you please change the PR title according [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) spec. It should have `:` after `feat(x)` -> `feat(x):`. Also please use `blockchain` as scope for the feature - `feat(blockchain):`. Also can you please give more details to the PR title. The PR title will be used as commit message after `Squash and merge` and will go directly to CHANGELOG

skambalin:
You should not change versions manually - it will be auto-detected during release, using to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) version detection.

PR #237, opened 2024-03-01
feat(ddc): Use internal SDK signers to reduce signature requests to web3 wallets

skambalin:
Use token delegation mechanism to delegate access to internal SDK signer. The real signer needs to sign only the delegation and the rest is signed by internal SDK signer.

Note: This optimization currently not applied to use cases with external auth tokens

### DIFF
--- a/packages/ddc/src/auth/sdkToken.ts
+++ b/packages/ddc/src/auth/sdkToken.ts
@@ -1,0 +1,46 @@
+import { Signer, Web3Signer, createRandomSigner } from '@cere-ddc-sdk/blockchain';
+
+import { AuthToken } from './AuthToken';
+
+const signerRegestry = new WeakMap<Signer, Map<string, Signer>>();
+
+const isWeb3Signer = (signer: Signer): signer is Web3Signer => {
+  return signer instanceof Web3Signer;
+};
+
+const getRegestry = (signer: Signer) => {
+  if (!signerRegestry.has(signer)) {
+    signerRegestry.set(signer, new Map());
+  }
+
+  return signerRegestry.get(signer)!;
+};
+
+const createSdkSigner = async (signer: Signer) => {
+  const randomSigner = createRandomSigner({ type: signer.type });
+  await randomSigner.isReady();
+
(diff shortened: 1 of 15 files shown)


packages/ddc/src/auth/sdkToken.ts
```diff
@@ -0,0 +1,46 @@
+import { Signer, Web3Signer, createRandomSigner } from '@cere-ddc-sdk/blockchain';
+
+import { AuthToken } from './AuthToken';
+
+const signerRegestry = new WeakMap<Signer, Map<string, Signer>>();
```
upalinski, line 5: typo in Regestry should be Regitstry (e -> i)

packages/ddc/src/routing/Router.ts
```diff
@@ -58,6 +60,7 @@ export class Router {
   async getNode(operation: RouterOperation, bucketId: BucketId, exclude: string[] = []) {
     this.logger.info('Getting node for operation "%s" in bucket %s', operation, bucketId);
 
+    this.sdkTokenPromise ||= createSdkToken(this.signer); // Request the token only once
```
upalinski, line 63: we create SDK token inside Router component, getNode method. Doesn't look great.

Can we do it during DdcClient initialisation? or some of low level APIs?

upalinski (approved):

packages/ddc/src/routing/Router.ts
skambalin: Router currently creates `StorageNode` instances and it is a common module which is used in both `FileStorage` and `DDCClient` packages. So it is the most convenient place to quickly put this token creation, but I agree that is not very logical place to do. Ideally we need to refactor the flow, so router responsibility was only to select nodes - not creating their instances. It also means that it will not create the tokens as well.

I tried to do a quick refactoring, but it leads to several potentially breaking changes, so I decided to postpone it and do in a separate PR. Added a [related task to backlog](https://www.notion.so/cere/Router-flow-refactor-20db02af412e44f980422ff56609e65f?pvs=4).

PR #239, opened 2024-03-27
feat: Re-try piece store operation from buffered content

skambalin:
- [x] Re-create piece with static content during re-try
- [x] Re-try file upload operations 

Note: Tests for this feature currently covers only happy pass - could not find an easy way to simulate failing storage node. Will try to find a way in another PR about re-triable download operations.


packages/ddc/src/constants.ts
```diff
@@ -3,6 +3,13 @@ import { GrpcStatus } from './grpc/status';
 export const KB = 1024;
 export const MB = 1024 * KB;
 
+/**
+ * Maximum size of a piece of content, in bytes.
```
upalinski, line 7: in variable name you have MIN. to either comment or variable name is incorrect

upalinski (approved):

packages/ddc/src/constants.ts
skambalin: Thanks - will fix

PR #257, opened 2024-07-10
Fix correlationId gRPC metadata key

jwoloch:


skambalin:
Duplicate of this change: https://github.com/Cerebellum-Network/cere-ddc-sdk-js/pull/258

PR #260, opened 2024-07-26
feature: Add CNS Record cache control option

skambalin:
Added a new option to all `read` operation of the SDK called `cacheControl`. Currently it allows passing only fixed value `no-cache` - to disable CNS cache completely, but later it can be extended if needed.

### DIFF
--- a/packages/ddc/src/nodes/NodeInterface.ts
+++ b/packages/ddc/src/nodes/NodeInterface.ts
@@ -16,6 +16,10 @@ type ActivityOptions = {
   correlationId?: string;
 };
 
+type CacheControlOptions = {
+  cacheControl?: 'no-cache';
+};
+
 /**
  * The `OperationAuthOptions` type defines the authentication options for a DDC operation.
  *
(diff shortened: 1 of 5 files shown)


mazhutoanton (approved):

packages/ddc/src/nodes/NodeInterface.ts
```diff
@@ -34,7 +38,8 @@ export type OperationAuthOptions = {
  * @hidden
  * @extends OperationAuthOptions
  */
-export type PieceReadOptions = ActivityOptions &
+export type PieceReadOptions = CacheControlOptions &
```
jwoloch, line 41: As discussed offline CacheControl is only implemented for CNS for now. You can pass it here, however it will not change the behavior. Content addressed with CID is immutable so no-cache can only degregade performance.

jwoloch: However we might extend CacheControl for other types of behaviour. Just FYI.

skambalin: The `readPiece` method of the `StorageNode` API has an option to read pieces by CNS name and under the hood it uses `resolveName` method which in tern uses `CNS API`. That is why the `cacheControl` option is added to read piece params, but at the end (low level) API it only used by `CNS get` method and is not added other GRPC calls.

jwoloch: Sure, I am resolving the comment.

jwoloch (approved):

PR #271, opened 2024-10-07
feature: Extend node operation errors with correlation ID and node URL

skambalin:
- Extend storage node RPC errors with `correlationId` and `nodeId` (gRPC endpoint)
- Make client router errors more informative instead of just `No nodes found...`


upalinski (approved):

tests/specs/Errors.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+      );
+    });
+  });
+
+  describe('Unreachable node', () => {
+    let client: DdcClient;
+    const bucketId = 1n;
+    const unreachableNode: RouterNode = {
+      mode: StorageNodeMode.Storage,
+      grpcUrl: `grpc://localhost:9099`,
```
khssnv, line 40: `localhost` required name resolution and may become IPv6 or something else unexpected. I suggest using `127.0.0.1` or `0.0.0.0`.

khssnv (approved):

PR #292, opened 2025-08-14
Feature/smart contract migration

mazhutoanton:


examples/node/smart-contract-usage/index.ts
```diff
@@ -0,0 +1,86 @@
+import { DdcClient } from '@cere-ddc-sdk/ddc-client';
+import { DEVNET } from '@cere-ddc-sdk/ddc';
+import { Blockchain } from '@cere-ddc-sdk/blockchain';
+
+async function main() {
+  console.log('🚀 Пример использования DDC Client со смарт-контрактами');
```
upalinski: is this file needed? if yes, please change comments and log messages to english.

also, make sure that this example style is aligned with other examples we have

examples/node/smart-contract-usage/README.md
```diff
@@ -0,0 +1,139 @@
+# DDC Client со смарт-контрактами
+
+Этот пример демонстрирует использование DDC Client с новой архитектурой на основе смарт-контрактов customer deposit.
```
upalinski: the same

packages/cli/src/deposit.ts
```diff
@@ -9,7 +9,7 @@ export type DepositOptions = {
 };
 
 export const deposit = async (client: DdcClient, amount: number, options: DepositOptions) => {
-  await client.depositBalance(options.clusterId, BigInt(amount * CERE), options);
+  await client.depositBalance(BigInt(amount * CERE));
```
upalinski, line 12: should there be clusterId param?

upalinski: if we have per cluster contracts (so there is no need to specify cluster id anymore), i wonder why in guides and examples we have 
```
// Депозит средств
await client.depositBalance(clusterId, amount);

```

mazhutoanton: no, the contract is deployed separately for each cluster

PR #295, opened 2025-12-09
migrate client sdk including sis into ddc-sdk

gharibyan:
1. migrated client sdk with sis initial implementation from activity sdk
2. fixed examples test run.


packages/client/src/types.ts
spoluyan, line 1: There are duplicate type definitions between `packages/client/src/types.ts` and `packages/client/src/sis/types.ts`. 
Both files define `ServiceUnavailableError`, `DataStream`, `Packet`, `HandshakeRequest`, etc. with different shapes:

- `types.ts`: `HandshakeRequest.type is 'publish' | 'subscribe' (string)`
- `sis/types.ts`: `HandshakeRequest.type is number (0 or 1)`

Consolidate all types into a single source of truth.

packages/client/src/client.ts
```diff
@@ -0,0 +1,116 @@
+import type { ClientConfig, SignedWallet } from './types';
+import { Client as SisClient } from './sis';
+import { ContextPath, Packet } from './sis/types';
+import Event from './event';
+import MPC from './mcp';
+import Wallet from './wallet';
+
+const DEFAULT_WALLET_CONFIG = 'hybrid label reunion only dawn maze asset draft cousin height flock nation';
```
spoluyan: Remove the default wallet entirely and require explicit wallet configuration, or throw an error if no wallet is provided.

packages/client/src/context.ts
```diff
@@ -0,0 +1,15 @@
+import { ContextInterface, ContextInput } from './types';
+
+export class ClientContext implements ContextInterface {
+  public agent_service: string;
+  public workspace: string;
+  public domain?: string;
```
spoluyan: Must be `stream`, not `domain`. Also, this field is required.

packages/client/src/client.ts
```diff
@@ -0,0 +1,116 @@
...
+      const stream = await this.sis.createStream(this.context as ContextPath, {});
+      return stream;
+    },
+    get: (streamId: string) => {
+      return this.sis.getStream(streamId);
+    },
+    publisher: async (streamId: string) => {
+      return this.sis.newPublisher(streamId);
+    },
+    subscribe: (streamId: string, callback: (data: { headers: Packet['headers']; data: any }) => void) => {
```
spoluyan: Potential memory leak.

Issues:

- The async IIFE is "fire-and-forget" with no way to unsubscribe
- No error handling if the subscription fails
- No return value to allow cleanup

packages/client/src/sis/http.ts
```diff
@@ -0,0 +1,179 @@
...
+            throw new StreamNotFoundError(streamMatch[1]);
+          }
+          throw new SISError(responseText || 'Not found', 'NOT_FOUND', 404);
+        }
+        if (responseText) {
+          throw new SISError(responseText || `HTTP ${response.status}`, 'HTTP_ERROR', response.status);
+        }
+        if (hasJson) {
+          try {
+            const j = await (response as any).json();
```
spoluyan: The code calls `response.text()` first, then later tries to call `response.json()`. Once the body is consumed by `text()`, `json()` will fail. This logic is convoluted and error-prone.

playground/src/BrowserStreaming.tsx
```diff
@@ -0,0 +1,561 @@
...
+  // Publisher state
+  const [publisherConnected, setPublisherConnected] = useState<boolean>(false);
+  const [messageInput, setMessageInput] = useState<string>('Hello from browser! 🌐');
+  const [sentCount, setSentCount] = useState<number>(0);
+  const [lastAck, setLastAck] = useState<number | null>(null);
+  const publisherRef = useRef<any>(null);
+
+  // Subscriber state
+  const [subscribed, setSubscribed] = useState<boolean>(false);
+  const [offsetInput, setOffsetInput] = useState<string>('');
```
spoluyan, line 67: The offset input field is rendered but never used when subscribing. Maybe file for Playground.

packages/client/src/sis/types.ts
```diff
@@ -0,0 +1,279 @@
...
+
+// =============================================================================
+// Error Types
+// =============================================================================
+
+/** Custom error for SIS operations */
+export class SISError extends Error {
+  constructor(
+    message: string,
+    public code?: string,
```
spoluyan, line 256: Minor - many places throw SISError without a code.

packages/client/src/client.ts
```diff
@@ -0,0 +1,116 @@
...
+      });
+      const json = await response.json();
+      if (response.ok) {
+        const resultData = json?.result?.data;
+        return resultData !== undefined ? resultData : json;
+      }
+      const errMessage = (json && json.error && json.error.message) || `Request failed with status ${response.status}`;
+      const error = new Error(errMessage);
+      // @ts-expect-error augment error with extra context
+      error.status = response.status;
```
spoluyan: Maybe it makes sense Instead of using @ts-expect-error, create a proper typed error class.

packages/client/src/sis/transport.ts
```diff
@@ -0,0 +1,452 @@
...
+  ) {
+    // Node.js environment
+    try {
+      // Dynamically import the WebTransport polyfill
+      // @ts-ignore - @fails-components/webtransport is an optional Node.js dependency
+      const { WebTransport, quicheLoaded } = await import('@fails-components/webtransport');
+      await quicheLoaded;
+      return WebTransport;
+    } catch (e) {
+      console.log(e);
```
spoluyan: `console.log(e);` leaks into production. Use proper logging infrastructure.

package.json
spoluyan, line 1: Not for this PR, but we need to fix Polkadot dependencies. The package declares dependencies on Polkadot utilities but doesn't specify them as peer dependencies, which could lead to duplicate versions. Maybe just add TODO for now.

packages/client/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "build:node": "microbundle --tsconfig tsconfig.build.json --format esm,cjs --target node",
+    "build:web": "microbundle --tsconfig tsconfig.build.json --format modern --output dist/browser.js",
+    "package": "clean-publish",
+    "clean": "rimraf dist package"
+  },
+  "dependencies": {
+    "@cere-activity-sdk/ciphers": "0.1.7",
+    "@cere-activity-sdk/signers": "0.1.7",
+    "@cere/embed-wallet": "^0.23.0",
+    "@fails-components/webtransport": "^1.4.8",
```
spoluyan: Listed twice (once in root `package.json`, once here). May cause version conflict in future.

packages/client/src/sis/client.ts
```diff
@@ -0,0 +1,474 @@
...
+  private initPromise: Promise<void> | null = null;
+
+  /**
+   * Creates a new unified SIS client.
+   *
+   * @param config - Client configuration with at least one bootstrap node
+   * @throws SISError if configuration is invalid
+   */
+  constructor(config: ClientConfig) {
+    this.config = {
```
spoluyan, line 69: Maybe move default values to constants?

packages/client/src/sis/client.ts
```diff
@@ -0,0 +1,474 @@
...
+  }
+
+  // ===========================================================================
+  // Cleanup
+  // ===========================================================================
+
+  /**
+   * Closes all connections and cleans up resources.
+   */
+  async close(): Promise<void> {
```
spoluyan: There's no way to gracefully shut down `ClientSdk`. The `SisClient` has a `close()` method but `ClientSdk` doesn't expose it.

spoluyan (changes requested):

packages/client/src/context.ts
gharibyan: Done!

packages/client/src/client.ts
gharibyan: Done.

package.json
gharibyan: added todo in readme

gharibyan:
@spoluyan tests are failing because of linter. if i run it at the moment there will be lot's of file changes. i'll fix that and testing flow with different pr.

packages/client/src/client.ts
```diff
@@ -0,0 +1,129 @@
...
+      const stream = await this.sis.createStream(this.context as ContextPath, {});
+      return stream;
+    },
+    get: (streamId: string) => {
+      return this.sis.getStream(streamId);
+    },
+    publisher: async (streamId: string) => {
+      return this.sis.newPublisher(streamId);
+    },
+    subscribe: (
```
spoluyan, line 77: Problems:

- The async IIFE runs independently with no cancellation mechanism
- Calling unsubscribe() doesn't stop the loop — it only closes the SIS client
- The loop continues even after unsubscribe() is called
- Multiple subscriptions create multiple uncancellable loops

Impact: memory leaks, resource waste, inability to stop subscriptions
Recommendation: return an AbortController or a cancellation function.

packages/client/src/client.ts
```diff
@@ -0,0 +1,129 @@
...
+            const headers = packet.headers;
+            const data = parsePacket(packet);
+            callback({ headers, data }, null);
+          }
+        } catch (error) {
+          callback(null, error as Error);
+        }
+      })();
+    },
+    unsubscribe: () => {
```
spoluyan: Problems:

- this.sis.close() closes all transports and HTTP clients
- If a user has both a publisher and subscriber, calling unsubscribe() will close both
- No way to unsubscribe from a single subscription without affecting other operations

Impact: breaks concurrent usage (publish + subscribe), unexpected connection drops
Recommendation: track individual subscriptions and close only the specific one

packages/client/src/sis/http.ts
```diff
@@ -0,0 +1,134 @@
...
+        body: body !== undefined ? JSON.stringify(body) : undefined,
+        signal: controller.signal,
+      });
+
+      if (!response.ok) {
+        if (response.status === 503) {
+          throw new ServiceUnavailableError();
+        }
+        if (response.status === 404) {
+          throw new SISError('Not found', 'NOT_FOUND', 404);
```
spoluyan: The HTTP client defines `StreamNotFoundError` but throws a generic SISError for 404

packages/client/src/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { Packet } from './sis/types';
+
+export const parsePacket = (packet: Packet) => {
+  const headers = packet.headers;
+  const contentType = headers?.['content-type'] || '';
+  let data: any;
+
+  if (contentType.includes('application/json')) {
+    const text = new TextDecoder().decode(packet.payload);
+    data = JSON.parse(text);
```
spoluyan: The first `JSON.parse` (line 10) can throw if the payload is invalid JSON. The second one (line 16) is wrapped in try/catch. Make both consistent.

packages/client/src/sis/transport.ts
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  /**
+   * Initializes the subscriber by opening a stream and sending the handshake
+   */
+  async init(): Promise<void> {
+    this.stream = await this.transport.createStream();
+    const writer = this.stream.writable.getWriter();
+    this.reader = this.stream.readable.getReader();
+    this.abortController = new AbortController();
```
spoluyan: Created but never used to abort operations. Either use it or remove it.

spoluyan (changes requested):

packages/client/src/sis/certificate.ts
spoluyan, line 1: SDK should not own logging; callers should plug in a logger.

packages/client/src/client.ts
```diff
@@ -0,0 +1,143 @@
...
+    },
+  };
+
+  public query = {
+    fetch: async (cubbyName: string, queryName: string, payload?: unknown, timeoutMs?: number): Promise<unknown> => {
+      const path = `agent-services/${this.context.agent_service}/cubbies/${cubbyName}/queries/${queryName}`;
+      const url = this.buildURL(this.agentRuntimeUrl, path);
+      const requestBody = {
+        params: payload,
+      } as CubbyQueryRequestBody;
```
spoluyan, line 126: The type says `payload` but the code sends `params`. The cast hides the mismatch.

packages/client/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export class CubbyError extends Error {
+  constructor(
+    message: string,
+    public status?: number,
+    public code?: string | number,
+  ) {
+    super(message);
+    this.name = 'CubbyRequestError';
```
spoluyan, line 41: `name` is set to `'CubbyRequestError'` while the class is `CubbyError`. That breaks the usual expectation that `err.name === 'CubbyError'` and can confuse logging/monitoring.

packages/client/src/client.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      if (timeoutMs) {
+        requestBody.timeoutMs = timeoutMs;
+      }
+      const response = await fetch(url, {
```
spoluyan: No `signal` or timeout. Long-running or stuck requests cannot be cancelled and can hang indefinitely.

spoluyan (changes requested):

PR #301, opened 2026-07-10
feat(ddc): Phase 2a — papi core foundation (stacked on #300)

spoluyan:
## Phase 2a — papi core foundation

**Stacked on #300 (Phase 1).** Base is `feat/ddc-protocol-compat-phase1`; review/merge #300 first. This is the first of four papi-migration sub-projects (2a core → 2b pallets → 2c ink! → 2d consumer cutover + breaking cutover).

Adds the papi-native chain-layer **foundation** to `@cere-ddc-sdk/blockchain`, **additively** — the legacy `@polkadot/api` API is untouched and stays the package's default; the papi core is reachable only via the new `./papi` subpath. The breaking cutover is 2d.

Design/plan: `docs/superpowers/specs/2026-07-10-papi-2a-core-design.md`, `docs/superpowers/plans/2026-07-10-papi-2a-core.md`.

### What's included
- **Descriptors** for devnet/testnet/mainnet, generated (`papi generate`) and shipped in-package via the `#descriptors` imports map + an `ensure-publishable-descriptors` step (handles the nested-package `npm pack` / `file:`-dep pitfalls).
- **`connect({network|wsUrl})`** → papi client + typed API, with a node/browser WS-provider split (`#ws-provider`) and a runtime **compat guard** (`assertCompatible`).
- **Signer layer** (`CereSigner`): `MnemonicSigner` (hdkd → `PolkadotSigner`), `ExtensionSigner` (papi pjs-signer, browser-guarded), `CereWalletSigner` (byte-level injection adapter — no `@cere/embed-wallet` runtime dep).
- **`./papi` subpath export** for the whole papi surface.
- **Dependency `overrides`** pinning the `@polkadot-api/*` set to `polkadot-api@2.1.8`'s versions — a legacy transitive (`@cere/embed-wallet-inject → extension-inject → substrate-connect → @polkadot-api/client@0.0.1-*`) otherwise skews it and breaks papi's `chainHead` client (`-32601`). Confirmed safe: that legacy `@polkadot-api/client` is a **light-client** path the wallet never invokes (cere-wallet-client only signs).
- **Native-ESM jest config** (`tests/jest.papi.config.ts`) for the papi live suite — `polkadot-api` is ESM-only and ts-jest's CJS down-compile breaks its `chainHead` client.

### Verification
- `check-types:ci`, `check-types`, `lint`, and the `packages/blockchain` build all green.
- Gated `papi.spec.ts` — **3/3 on devnet**: `connect` + typed read of `ClustersGovParams` (snake_case) + a **real finalized `System.remark`** signed by `MnemonicSigner`.
- Default `npm test` path unaffected (papi suite is env-gated + on its own ESM config).

### Deferred to later sub-projects (documented)
- **testnet/mainnet:** their public RPCs don't yet implement papi's new JSON-RPC (chainHead) — devnet does. Needs a version-matched `@polkadot-api/polkadot-sdk-compat` shim (latest 2.4.1 mismatches papi 2.1.8's provider format). Tracked.
- 2b pallets · 2c ink! (via `@polkadot-api/ink-contracts` + `ContractsApi`, since sdk-ink is pallet-revive-only) · 2d consumer cutover + breaking major.
- Minor polish: `assertCompatible` test + drop `as any`; `MnemonicSigner` soft-only derivation; `connect(string)` unknown-host → mainnet default.


spoluyan:
I reviewed this in a throwaway worktree (didn't touch the working branch). The papi plumbing itself is solid and I verified the signer/ws/descriptor wiring against the installed papi + hdkd type defs — those are correct. But there are two things I think need to land before merge.

## 🔴 Blocker — branch isn't based on its own declared base

The PR targets `feat/ddc-protocol-compat-phase1` (tip `350d10d`), but the merge-base is `3c35a0f`. The base branch's three most recent commits — `350d10d` (`continue-on-error`), `334f2b7`, `4c3878d` — are **not in the PR head**.

The knock-on effect is serious: CI now runs **zero real assertions**. The last commit rewrites `tests/package.json`'s `test` script to run only `papi.spec.ts`, which `describe.skip`s without `CERE_CHAIN_TESTS`. The other 12 specs — including `ChainCompat.spec.ts` and `metadata-audit.spec.ts`, which base commit `350d10d` explicitly called the *blocking gates* — no longer run at all. CI is green because it asserts nothing. The PR body's "Default `npm test` path unaffected" is contradicted by this commit.

Also the workflow on the PR head is missing the base's `continue-on-error: true`, so that intent is silently lost too.

**Action:** rebase onto current `feat/ddc-protocol-compat-phase1`, and make a deliberate call on the test script — at minimum the chain-compat unit specs need to keep running. (Happy to do this if you want.)

## 🟡 `connect(string)` → mainnet default is a confusing failure mode

`inferNetwork()` returns `'mainnet'` for any unknown host. But mainnet's public RPC doesn't yet implement papi's `chainHead` RPC (documented in the code), so `connect('wss://something-internal')` silently targets mainnet and then dies at runtime with `RpcError -32601`. A silent network misroute that then fails with a cryptic RPC error is exactly the footgun this layer exists to prevent. Either throw on an unrecognized host or make devnet/testnet explicit-only.

## 🟡 `assertCompatible` throws on a typo (fail-closed, but misleading)

`client.ts` does `(statics.compat.tx as any)[pallet]?.[call]`. `compat.tx` is actually strongly typed (`CompatHelperApi<TxFromPalletsDef<...>>`), so the `as any` discards real type safety. Behavior is fail-closed (misspelled pallet → `entry` undefined → throws "not compatible"), but the message is wrong for a not-found/typo case. Drop `as any`, let the descriptors type the lookup, and throw a distinct "unknown call" error when absent.

## ✅ Verified correct (against papi 2.1.8 + hdkd type defs)

- **`MnemonicSigner`** — I initially suspected a `KeyPair`/`sign` shape mismatch, but `hdkd-helpers`' `KeyPair` *is* `{ publicKey, sign: (msg: Hex) => Uint8Array }`, and `getPolkadotSigner` wants `(input: Uint8Array) => Uint8Array`. `Hex = Uint8Array | string`, so it's type- and behavior-safe. Hard/soft re-gluing matches keyring URI semantics.
- **`ExtensionSigner` / `CereWalletSigner`** — clean adapters. Not importing `@cere/embed-wallet` (its `signMessage(string): Promise<string>` doesn't fit raw-bytes `getPolkadotSigner`) is the right call — leaves encoding to the caller.
- **`ws-provider` node/browser split** via the `#ws-provider` imports map is consistent: tsc resolves the node variant for types, bundlers pick the browser condition. Node `ws` (net/tls/crypto) never reaches a browser bundle.
- **`ensure-publishable-descriptors.mjs`** — correctly handles the nested-package npm-packlist trap and strips `file:.papi/descriptors` so consumers don't hit `ERR_PNPM_LINKED_PKG_DIR_NOT_FOUND`. Idempotent.
- **Metadata `.scale` committed** → `papi generate` in `prepare` is offline/deterministic; the Node 18→22 bump is genuinely required (papi CLI import attributes).

## Recommendation

**Request changes.** The papi foundation is well-built, but the rebase gap + the test-script change would silently disable the repo's test gate. Once those are sorted this is a strong base for 2b.

spoluyan (commented):
Marking this as a comment review (GitHub won't let the PR author formally 'request changes' on their own PR). Please treat the detailed comment above as a request-changes — primarily the rebase gap that's making CI assert nothing.

tests/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Cere DDC SDK tests",
   "private": true,
   "scripts": {
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --config jest.papi.config.ts --runInBand",
```
yahortsaryk: 🔴 **This drops the existing test suite from CI.** This repoints the package's default `test` to the papi-only config, whose `testMatch` is just `specs/papi.spec.ts` (and that spec is `describe.skip` without `CERE_CHAIN_TESTS`). Since CI runs `test:ci` → `lerna run test` → this script, the whole existing suite (`Blockchain`, `DdcClient`, `StorageNode`, `Auth`, `Routing`, `ChainCompat`, `metadata-audit`, and the docker `globalSetup`) no longer runs at all.

This also contradicts the PR description ("Default `npm test` path unaffected"). Suggest adding a **new** script instead, e.g. `"test:papi": "NODE_OPTIONS=--experimental-vm-modules jest --config jest.papi.config.ts --runInBand"`, and leaving `"test": "jest"` intact.

packages/blockchain/src/papi/client.ts
```diff
@@ -0,0 +1,86 @@
...
+  // can't flip the network; fall back to the raw string if it doesn't parse.
+  let host = wsUrl;
+  try {
+    host = new URL(wsUrl).hostname;
+  } catch {
+    /* not a parseable URL — scan the raw string */
+  }
+  if (host.includes('devnet')) return 'devnet';
+  if (host.includes('testnet')) return 'testnet';
+  return 'mainnet';
```
yahortsaryk, line 69: 🟡 **Risky default.** An unrecognized host falls through to `mainnet` — the one network with real funds, and (per this PR's own notes) the one papi doesn't even work against yet. A typo'd endpoint silently binds to mainnet descriptors. I'd prefer throwing on an unknown host, or defaulting to `devnet`. (I see it's listed as deferred polish, but defaulting to the most dangerous network seems worth fixing before this ships.)

packages/blockchain/src/papi/client.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export function connect(opts: ConnectOptions | string): CereClient {
+  const network = typeof opts === 'string' ? inferNetwork(opts) : opts.network;
+  const wsUrl = typeof opts === 'string' ? opts : (opts.wsUrl ?? CERE_WS[opts.network]);
+
+  // NOTE: Cere's public testnet/mainnet RPC endpoints don't yet implement papi's new
+  // JSON-RPC (chainHead_v1/…) — devnet does. Connecting to those needs a version-matched
+  // `@polkadot-api/polkadot-sdk-compat` shim (latest 2.4.1 mismatches papi 2.1.8's provider
+  // message format); tracked as a follow-up. devnet connects natively.
+  const client: PolkadotClient = createClient(getWsProvider(wsUrl));
```
yahortsaryk, line 86: Since testnet/mainnet RPCs don't implement papi's chainHead yet, `connect({ network: 'testnet' | 'mainnet' })` builds a client that fails later with a cryptic `-32601`. Consider guarding here and throwing `ChainIncompatibleError` ("papi not yet supported on {network}") so the failure is clear at the call site rather than deep in a query. Documented as deferred — just flagging the UX.

yahortsaryk (commented):
Solid additive foundation — the `./papi` isolation is the right call for a staged migration, and the rationale comments (ensure-publishable-descriptors, the ws-provider split, the overrides pin) are excellent. Granular comments inline; one is blocking-worthy (the default `test` script).

package.json
```diff
@@ -57,5 +57,37 @@
     "nohoist": [
       "playground/**"
     ]
+  },
+  "comments": {
+    "overrides": "polkadot-api's @polkadot-api/* packages are version-locked. A legacy transitive (@cere/embed-wallet-inject -> @polkadot/extension-inject -> @substrate/connect -> @polkadot-api/client@0.0.1-*) drags in ancient @polkadot-api/* that hoist and skew against polkadot-api@2.1.8, breaking its chainHead client (RpcError -32601). Pin the shared set to polkadot-api@2.1.8's versions."
+  },
+  "overrides": {
```
yahortsaryk, line 64: Nit / maintenance: these ~26 `@polkadot-api/*` entries are exact-pinned and apply repo-wide, so any future `polkadot-api` bump has to resync all of them by hand. Worth a renovate group or a CI check to catch drift. (Also `@polkadot-api/polkadot-sdk-compat` is pinned to `2.4.1` here, which the PR body calls the mismatched version — harmless since it's unused until 2b, just noting.)

packages/blockchain/src/papi/signers/MnemonicSigner.ts
```diff
@@ -0,0 +1,45 @@
...
+  readonly address: string;
+  private readonly signer: PolkadotSigner;
+
+  constructor(uri: string, opts: { type?: 'sr25519' | 'ed25519' } = {}) {
+    // Split `//hard/soft` derivation off the phrase, mirroring keyring URIs.
+    const [phrase, ...paths] = uri.split('//');
+    const trimmedPhrase = phrase.trim();
+
+    if (!trimmedPhrase) {
+      throw new Error('MnemonicSigner: empty mnemonic/URI');
```
yahortsaryk: Rejecting an empty phrase is a good footgun guard 👍 — note it also makes `//Alice`-style dev URIs (empty base phrase) throw, which differs from `@polkadot/keyring`'s `addFromUri`. Fine if intentional. Separately, the `//hard/soft` derivation-path reconstruction is untested (only a plain-seed `System.remark` is covered) — worth a unit test given the "soft-only derivation" caveat in the description.

packages/blockchain/src/papi/client.ts
spoluyan: Agreed this is the wrong default, and it's worth fixing at the source rather than in each consumer (`DdcClient`/`FileStorage` both funnel a string through `connect()`, so the same typo footgun surfaces in #305).

Not changed in this pass because `inferNetwork` only picks descriptor sets, and since this review mainnet's RPC was upgraded — so the "papi doesn't work against mainnet at all" half of the risk is gone, which lowers the severity but not the wrongness. Throwing on an unrecognized host is the right behavior; tracking it as a follow-up rather than a mid-stack change to shared connect() semantics.

tests/package.json
spoluyan: Resolved over the course of the stack: by 2d-iii the papi suite carries offline unit specs that run unconditionally, so `test:ci` executes **26 passing tests / 8 suites** with `CERE_CHAIN_TESTS` unset. The docker harness (and the specs that depended on it) was retired deliberately in 2d-iii, documented in `tests.yaml` and MIGRATION.md rather than left implicit. The "default `npm test` unaffected" line in this PR's description was wrong at the time — accurate now only in the sense that `npm test` runs the surviving suite.

package.json
spoluyan: Noted — the ~26 exact-pinned `@polkadot-api/*` overrides do have to be resynced by hand on a `polkadot-api` bump. They exist to stop npm resolving mismatched transitive copies of the papi codec tree (which produced runtime decode failures). A renovate group or a CI drift check is the right long-term answer; tracked as a follow-up.

packages/blockchain/src/papi/signers/MnemonicSigner.ts
spoluyan: Correct that rejecting an empty phrase also rejects bare `//Alice`-style dev URIs, unlike `@polkadot/keyring`'s `addFromUri` — that's intentional (silently substituting the well-known dev phrase is the footgun being removed). I hit it myself writing a spike and the fix was to pass the full dev mnemonic explicitly, which is the behavior we want.

On the untested `//hard/soft` derivation: fair, and still open. The offline signer specs cover plain-seed sr25519/ed25519 sign+verify round-trips but not path derivation; adding a derived-path vector is a cheap follow-up.

packages/blockchain/src/papi/client.ts
spoluyan: This one has since been resolved by the chains rather than by a guard: all three networks (devnet, testnet, mainnet) now serve papi'\''s stabilized `chainHead_v1` JSON-RPC, so `connect({ network: '\''testnet'\'' | '\''mainnet'\'' })` connects natively and the cryptic `-32601` path no longer exists. Mainnet'\''s RPC was upgraded after this review; I verified all three live this week (chain reads + writes work on each, and a full funded write round-trip passed on testnet).

The comment at `client.ts` on the final tip reflects that, and `@polkadot-api/polkadot-sdk-compat` is no longer needed. Adding a `ChainIncompatibleError` guard for '\''papi not supported on {network}'\'' would now be dead code — the remaining unknown-host risk is the `inferNetwork` mainnet default, which is tracked in the thread above.

yahortsaryk (approved):
Approving. Comments addressed across the stack: the default test-script coverage concern is resolved by 2d-iii (offline unit suite runs unconditionally), the testnet/mainnet chainHead concern is moot (public RPCs upgraded, compat shim dropped), and the remaining items are by-design (MnemonicSigner empty-phrase) or deferred non-severe follow-ups (inferNetwork unknown→mainnet footgun, overrides maintenance). No unresolved severe runtime issue in this PR's scope.

PR #302, opened 2026-07-10
feat(ddc): Phase 2b — papi-native pallet layer (stacked on #301)

spoluyan:
## Phase 2b — papi-native pallet layer (stacked on #301)

Third sub-project of the papi migration (2a core → **2b pallets** → 2c ink!/customers → 2d cutover). Builds the read/write pallet surface for the four non-customer pallets on top of the Phase 2a papi client, **additively** — the legacy `@polkadot/api` chain layer and the package's `.` public API are untouched. Base branch is `feat/papi-2a-core` (#301).

### What's added (all under `src/papi/`)
- **`client.tx`** — `send`/`batchSend`/`batchAllSend`/`sudo`/`sudoAs` + event extraction into the existing `{ section, method, data, payload }` shape; decoded `Pallet.Error` on dispatch failure.
- **`client.chain`** — `getCurrentBlockNumber`/`getAccountFreeBalance`/`getNextNonce`/`getChainDecimals`/`formatBalance`.
- **`client.clusters` / `client.clustersGov` / `client.nodes` / `client.staking`** — the four pallets' read/write methods, method-parity with the legacy `*Pallet` classes.
- **`pallets/mapping.ts`** — single home for the snake_case↔camelCase translation (reads → existing `types.ts` camelCase domain shapes; writes → papi tx builders).
- **`CereApi`/`Sendable`** types.

### Design decisions
- **Reads re-map to the existing camelCase domain types**; **writes return papi tx builders** submitted via `client.tx.send`.
- **Customers pallet deferred to 2c** (buckets/ledger + the ink! deposit contract).
- **No metadata-audit port** — papi's typed descriptors + 2a's `assertCompatible` replace it.
- **Write verification = encode + `assertCompatible`** against the live devnet runtime (no state-changing governance submits); the `send`→finalized→events path is proven with a benign `System.remark`.

### Verification
- Devnet-gated ESM suite (`jest.papi.config.ts`): `papi-{tx,chain,clusters,clustersGov,nodes,staking}.spec.ts` green against live devnet; reads decode into the domain shapes, all mutating builders encode + pass `assertCompatible`.
- Gates green: `check-types:ci`, `check-types`, `lint`, full `npm run build` (6 projects). Default `npm test` path unaffected (papi suite is env-gated + separate ESM config).
- Every pallet's decoded field names + call arg shapes were probe-verified against live devnet and cross-checked against the generated descriptors (not guessed).

### Notable chain-reality corrections found during implementation
- `treasury/validators/clusterReserve` shares are `Perquintill`/u64 → must encode as `bigint` (papi value-compat rejects `number`).
- Mainnet runs an **older** gov-params shape (`unit_per_*`, no gpu/cpu/ram, no `customer_deposit_contract`); reads are cross-shape-tolerant, mainnet **writes** deferred to the compat backlog (2b targets devnet).
- `DdcNodes.StorageNodes` reads by a **bare** key but writes **enum-wrap** it; host/domain are raw bytes (`Binary.toText`/`fromText`).
- **No `CDNs` storage** on any runtime → the two legacy CDN staking read methods are dropped (public-parity break to note in the 2d migration guide).

### Deferred to 2d / backlog
- CI job running the gated papi encode suite against devnet (the `as any` tx-arg casts mean `check-types` can't catch field-name regressions; the encode+compat suite is the net).
- Extend `StorageNodeMode` with the runtime's `Compute` variant; testnet/mainnet write compat shim; serialize live-tx specs to kill the cross-file nonce/teardown flake; `StakingLedger.unlocking` first-class typing; a `tx.send` dispatch-failure negative test; 2d migration-guide note for the dropped CDN methods.


spoluyan (commented):
(superseded — full summary + inline notes are on review #4756862144 above/below)

tests/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Cere DDC SDK tests",
   "private": true,
   "scripts": {
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --config jest.papi.config.ts --runInBand",
```
spoluyan: 🔴 **CI "Run integration tests" now runs zero tests.**

This repoints the `tests` package `test` script to the papi-only config (`jest.papi.config.ts`), whose specs are gated by `process.env.CERE_CHAIN_TESTS ? describe : describe.skip`. That var is **not set anywhere in `.github/`**, so every spec self-skips.

And the `tests` package is the **only** package with a `test` script (no other package defines one), so the full CI path collapses: `tests.yaml:55 npm run test:ci` → `lerna run test` → only this package → papi suite → **all skipped**.

The docker E2E/integration suite (`tests/specs/**/*.spec.ts` via `jest.config.ts`) is dropped entirely, with no remaining script invoking it. The motivation is sound (flaky docker E2E + ESM-only papi specs breaking under CJS ts-jest), but a green CI that runs zero tests hides the regression.

Suggested: add `test:unit` (papi config) + `test:e2e` (docker config) and have `test` run both, so `lerna run test` still exercises the integration suite. If docker E2E is being deliberately retired, document the plan and wire the papi suite to actually run against a devnet (set `CERE_CHAIN_TESTS` in CI).

packages/blockchain/src/papi/tx.ts
```diff
@@ -0,0 +1,75 @@
...
+    section: e.type,
+    method: e.value?.type,
+    data: e.value?.value,
+    payload: e.value?.value,
+  }));
+}
+
+export function createTxApi(api: CereApi): TxApi {
+  const send = async (tx: Sendable, { signer, nonce }: SendOptions): Promise<SendResult> => {
+    const res = await tx.signAndSubmit(resolveSigner(signer), nonce == null ? undefined : { nonce });
```
spoluyan, line 55: 🟡 Here `data` and `payload` are set to the *same* value (`e.value?.value`). The legacy path (Blockchain.ts:194-202) builds a real keyed `payload`: `{ fieldName: value }` from `event.meta.fields`. So a consumer reading `event.payload.someField` gets nothing here, even though the `Event` type advertises `payload: Record<string, any>`.

Either reconstruct the keyed payload from papi's decoded event value, or drop `payload` from the type and note it as a deliberate 2b gap.

packages/blockchain/src/papi/pallets/nodes.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    async listStorageNodes() {
+      const entries = await api.query.DdcNodes.StorageNodes.getEntries();
+      return entries.map((e) => toStorageNode(e.value)).filter((n): n is StorageNode => n != null);
+    },
+    setStorageNodeProps(key, props) {
+      return api.tx.DdcNodes.set_node_params({
+        node_pub_key: storagePubKey(key),
+        node_params: buildStorageNodeParams(props),
+      } as any) as Sendable;
```
spoluyan, line 43: 🟡 This filter is dead — `toStorageNode` always returns a non-null object, and would throw on `value.props` before reaching the filter if `e.value` were null. If the intent is guarding a null storage entry, push it inside `toStorageNode` (`if (!value) return null`) where it can actually fire.

packages/blockchain/src/papi/pallets/nodes.ts
```diff
@@ -0,0 +1,44 @@
...
+    createStorageNode(key, props) {
+      return api.tx.DdcNodes.create_node({
+        node_pub_key: storagePubKey(key),
+        node_params: buildStorageNodeParams({ domain: '', ssl: false, ...props }),
+      } as any) as Sendable;
+    },
+    // `DdcNodes.StorageNodes` is keyed by a BARE SS58 string, not the
+    // `NodePubKey` enum (verified live — see `toStorageNode`'s doc comment
+    // in mapping.ts): pass `key` directly, unlike the enum-wrapped
+    // `node_pub_key` TX arg below.
```
spoluyan, line 27: 🟢 Nit: `buildStorageNodeParams` already defaults `domain ?? ''` and `ssl ?? false` (mapping.ts:200-202), so the spread here is redundant — and inconsistent with `setStorageNodeProps` a few lines down, which doesn't do it. Pick one spot for the defaults.

spoluyan (commented):
## Review — Phase 2b papi-native pallet layer

Strong, carefully-done work. `mapping.ts` is the heart of the change and it's done right — every snake_case↔camelCase translation is probe-verified against live devnet and cited with descriptor hashes. Method parity with the legacy `*Pallet` classes is complete (clusters, clustersGov, nodes, staking), and the additive design keeps the legacy `@polkadot/api` surface untouched.

**One blocking issue before merge:** a CI commit (`aa03c50`) silently reduces the integration-tests step to running **zero** tests. Details in the inline note on `tests/package.json`.

---

### 🔴 Blocking

**CI "Run integration tests" now runs nothing.** `aa03c50` repoints the `tests` package `test` script to the papi-only config (`jest.papi.config.ts`), whose specs are gated by `process.env.CERE_CHAIN_TESTS ? describe : describe.skip`. That var is **not set anywhere in `.github/`**, so every spec self-skips. And the `tests` package is the **only** package with a `test` script — no other package defines one. Tracing CI: `tests.yaml:55` → `npm run test:ci` → `lerna run test` → only the `tests` package → papi suite → **all skipped**. The entire docker E2E/integration suite (`tests/specs/**/*.spec.ts` via `jest.config.ts`) is dropped, and there's no remaining script that invokes it. A green CI that runs zero tests is worse than a red one. Recommend either keeping both suites under `test:ci` (a `test:unit` + `test:e2e` aggregator), or — if docker E2E is being deliberately sunset — documenting the plan and wiring the papi suite to actually run against a devnet.

### 🟡 Should address

- **`shapeEvents` discards the structured `payload`** (inline note on `tx.ts`). Legacy builds `{ fieldName: value }`; papi sets `data` and `payload` to the same raw value. The `Event` type advertises a `payload` it doesn't deliver.
- **`toStorageNode`'s `.filter(n => n != null)` is dead** (inline note on `nodes.ts`). The mapper always returns non-null; a null guard belongs inside `toStorageNode`.

### 🟢 Nits

- `createStorageNode` double-defaults `domain`/`ssl` (inline note on `nodes.ts`) — `buildStorageNodeParams` already does it; inconsistent with `setStorageNodeProps`.
- `decimalsCache` is the only mutable closure state in the layer — a one-liner that it's intentionally cached per-client would stop a later reader from "fixing" the capture.

### What's good

- `mapping.ts` cross-shape read tolerance for mainnet's older `unit_per_*` protocol-params naming — graceful fallback to `undefined` instead of `BigInt(undefined)`.
- The `Perquintill`→`bigint` write-path gotcha is caught and documented with the actual papi error string — `assertCompatible` won't catch it, so the explicit `BigInt()` on writes is load-bearing.
- Honest scoping: dropped CDN reads, deferred `Compute` mode variant, mainnet write gap, missing CI devnet job are all called out rather than hidden.
- Encode + `assertCompatible` (no submit) is a reasonable net for field-name regressions given the `as any` tx-arg casts defeat `check-types` — and the PR body says this explicitly.

**Question:** is the docker integration suite being deliberately retired here/2d, or should `aa03c50` keep it in CI? Determines whether the blocker is "fix the script" or "write the migration note now."

---
*Reviewed against base `feat/papi-2a-core`. Current branch `feat/single-cluster-config` left untouched; used a temporary worktree for the read.*

tests/package.json
spoluyan: Outdated against the current branch — verified. By 2d-iii the papi suite includes offline unit specs (endpoint-resolver, retry, ddcclient-config, barrel, resolve-client, metadata-audit, …) that run unconditionally, so `test:ci` now runs **24 passing tests across 8 suites** even with `CERE_CHAIN_TESTS` unset. The docker-E2E sunset is documented inline in `tests.yaml` and the integration step is `continue-on-error` by design (deliberate unit-only CI); the chain-gated suites run locally with `CERE_CHAIN_TESTS=1`.

packages/blockchain/src/papi/tx.ts
spoluyan: Not a defect — verified against consumers. `extractCreatedBucketIds` reads `e.payload?.bucket_id`, and `shapeEvents` sets `payload = e.value.value`, which for `BucketCreated` is the decoded `{ bucket_id, … }` object — so `payload.bucket_id` resolves. Confirmed by the live testnet round-trip (`createBucket` returns the new bucket id). The legacy keyed `{fieldName: value}` payload isn't required because the extractors were written against papi's decoded value.

packages/blockchain/src/papi/pallets/nodes.ts
spoluyan: Fixed — removed the dead `.filter((n): n is StorageNode => n != null)`. `getEntries` values are always present and `toStorageNode` returns non-null (it throws on a malformed value rather than returning null), so the filter could never fire.

packages/blockchain/src/papi/pallets/nodes.ts
spoluyan: Fixed — `createStorageNode` now passes `props` straight to `buildStorageNodeParams` (which already defaults `domain ?? ''` / `ssl ?? false`), matching `setStorageNodeProps`.

PR #303, opened 2026-07-10
feat(ddc): Phase 2c — papi Customers pallet + ink! deposit contract (stacked on #302)

spoluyan:
## Phase 2c — papi Customers pallet + ink! customer-deposit contract (stacked on #302)

Fourth sub-project of the papi migration (2a core → 2b pallets → **2c ink!/customers** → 2d cutover). Adds the Customers pallet to the papi client, **additively** under `src/papi/` — the legacy `@polkadot/api` chain layer and the package `.` public API are untouched. Base branch is `feat/papi-2b-pallets` (#302). This closes **goal #2** of the modernization: the customer ledger that moved from a pallet to a smart contract.

### What's added
- **`client.customers`** — buckets (`getBucket`/`getBucketsCount`/`listBuckets`/`createBucket`/`setBucketParams`/`removeBuckets` + `extractCreated/RemovedBucketIds`) on the `DdcCustomers` pallet, and the deposit/balance surface (`getStackingInfo`, `deposit`/`depositExtra`/`depositFor`/`unlockDeposit`/`withdrawUnlockedDeposit`).
- **`contracts/customerDeposit.ts`** — the ink! layer via `@polkadot-api/ink-contracts` (dynamic codec over the shipped ABI) + `ContractsApi.call` (dry-run reads/gas) + `Contracts.call` (tx). Per-cluster address resolution from `clustersGovParams.customerDepositContract`, gas sizing (dry-run ×2 margin, generous-ceiling fallback, capped at `max_block`).
- **`mapping.ts`** +`toStakingInfo`/`toBucket`.

### Design
- **Deposit/balance route contract-first** (a cluster with a real `customerDepositContract`), with a **new-runtime pallet fallback**; mainnet's legacy pallet shapes are deferred to 2d.
- Reads re-map to the existing camelCase `types.ts` shapes; a failed balance dry-run **throws** (never conflated with "no deposit").
- Contract *deployment* (instantiate/upload) is out of scope.

### Verification (live devnet)
- **Real funded-seed deposit round-trip** proves the contract path end-to-end: the test asserts the built tx is a `Contracts.call` (`decodedCall`) and that the on-chain balance increases by the deposited amount. Balance reads, buckets, and pallet-fallback encode + `assertCompatible` are also covered.
- Full papi suite passes serially (`--runInBand`, 26/26 across 10 specs); gates green: `check-types:ci`, `lint`, full `npm run build`. Default `npm test` path unaffected (papi suite env-gated + separate ESM config).

### Chain-reality findings (spike + live-verified)
- ink! decode is **two-level**: dispatch `{success,value:{flags,data}}` then `value.dec(data)` → `{success,value:Option<Ledger>}`; raw snake_case ABI labels (`DdcBalancesDepositor::deposit`, etc.); `System.BlockWeights` is async; `Contracts.call` dest is a `MultiAddress`.
- `ExistentialDeposit` = 10 CERE (min deposit); repeat contract deposit is a **top-up** (no `AlreadyPaired` on the contract path).
- `deposit_for` pallet args are `{owner,cluster_id,value}` and, like the `{cluster_id}`-keyed `withdraw_unlocked_deposit` + `ClusterLedger`, are not on the mainnet static type (reached via `as any`, same pattern as 2b).
- **`readBalance` overrides the decoded `Ledger.owner`** with the queried account: the contract returns a fixed wrong `owner` on-chain (a documented follow-up), while `total`/`active` are correct — so the SDK returns the authoritative queried owner.

### Deferred to 2d / backlog
- Root-cause the contract `Ledger.owner` on-chain anomaly (currently masked by the queried-owner override).
- Live contract-path smoke tests for `depositFor`/`unlockDeposit`/`withdrawUnlockedDeposit` (statically verified + fallback-compat-checked today; only `deposit`/`depositExtra` get live contract coverage).
- Mainnet legacy pallet shapes + testnet/mainnet JSON-RPC compat shim; the `DdcClient`/`cli` cutover, legacy-code removal, new major, and migration guide.


spoluyan (commented):
# Phase 2c — papi Customers pallet + ink! deposit contract

**Verdict: Approve with concerns.** Well-architected, additive, and verified against a live devnet round-trip. One packaging issue needs verification before merge; the rest are minor/parity notes. (Would have filed as "request changes" for #1, but GitHub forbids that on your own PR — so please verify #1 manually before merge.)

## Strengths

- **Genuinely additive.** Everything lands under `src/papi/`; the legacy `@polkadot/api` layer and the package `.` public API are untouched. `getStackingInfoLegacy` (unused downstream — zero consumers outside its own file) is correctly dropped.
- **Contract-first-with-fallback is the right shape.** `contractOrPallet` DRYs all five deposit/withdraw methods cleanly; each differs only in ink message/value/args and the pallet thunk.
- **`readBalance` correctness property is sound and well-justified.** Dispatch failure and ink `!success` both **throw**; only `Ok(None)` returns `undefined`. This is the crucial distinction — a broken query must never read as a zero balance. The comments explain the two-level decode (`dispatch → ink MessageResult → Option<Ledger>`) clearly.
- **Gas sizing is thoughtful.** 2× margin over `gas_required`, generous-ceiling fallback on any dry-run failure, capped at `max_block`, and — nicely — the `maxBlock` promise memo is cleared on rejection so a transient fetch doesn't permanently poison `estimateGas`.
- **The live deposit round-trip is strong evidence.** Asserting `decodedCall.type === 'Contracts'` / `.value.type === 'call'` is exactly the right guard — it proves the tx actually routes through ink! rather than silently falling back to the pallet (both move the same ledger, so the balance delta alone couldn't tell them apart).
- **Comments are excellent throughout** — they explain *why*, cite live findings, and correct earlier wrong assumptions (the `AlreadyPaired` note, the SS58-vs-hex zero-address sentinel).

## Concerns

### 1. ⚠️ [Blocker candidate — verify before merge] The ABI JSON likely won't resolve in the published artifact

`build:papi` is plain `tsc -p tsconfig.papi.json`. With `resolveJsonModule: true`, tsc type-checks `import abi from '../../customer_deposit.json'` but **does not copy the JSON into `outDir`** — tsc only emits `.js`/`.d.ts` for compiled `.ts` files. The blockchain `package.json` `files` field ships only `dist`, `.papi`, `package` — so `src/customer_deposit.json` (27 KB) is **not** in the published tarball.

Emitted JS from `dist/papi/contracts/customerDeposit.js` will resolve `../../customer_deposit.json` → `dist/customer_deposit.json`, which won't exist. Risk: `ERR_MODULE_NOT_FOUND` the first time a consumer calls any deposit/balance method on the papi entry.

The "full `npm run build` passes" gate doesn't catch this — tsc compiles cleanly because the JSON *is* present in `src/` at compile time; only runtime resolution in the packed artifact breaks. And the contract path is env-gated (`CERE_CHAIN_TESTS`), so CI skips it too.

**Verify:** `npm run build && npm pack` in `packages/blockchain`, then check whether `dist/customer_deposit.json` exists. **Fix if missing:** add a copy step to `build:papi` (`cpy`/`cp`), or inline the ABI as a `.ts` literal, or add `src/customer_deposit.json` to `files`.

### 2. [Parity] `readBalance` owner override diverges from legacy

Legacy `CustomerDepositContract.readBalance` returns the contract's decoded (broken) `owner`. The new code overrides it with the queried `accountId`:

```js
return { ...toStakingInfo(ledger), owner };
```

The PR body documents *why* (the contract returns a fixed wrong owner on-chain; `total`/`active` are correct). This is arguably a **fix**, not a regression — but it's a behavioral change vs. the legacy path, so anything (unlikely) relying on the broken value would see different output. The override is the right call; just flagging that it's intentional and should be noted in the 2d migration guide.

Root-causing the contract's bogus `owner` is correctly deferred to 2d.

### 3. [Consistency] Owner handling differs between contract path and pallet fallback

In `getStackingInfo`, the contract path overrides `owner` with the queried account (via `readBalance`). The pallet-fallback path does **not** — it passes the raw `ClusterLedger` value to `toStakingInfo`, which trusts `value.owner`:

```js
const value: any = await (api.query.DdcCustomers as any).ClusterLedger.getValue(clusterId, accountId);
return value == null ? undefined : toStakingInfo(value);
```

If `ClusterLedger` storage carries no `owner` field (or a stale one), the fallback returns a different/undefined owner than the contract path for the same account. Worth confirming `ClusterLedger` actually has a correct `owner`, or apply the same queried-account override for symmetry.

### 4. [Test scope] Three of five deposit methods lack live contract coverage

Only `deposit`/`depositExtra` get the live round-trip; `depositFor`, `unlockDeposit`, `withdrawUnlockedDeposit` are statically encoded + fallback-compat-checked. The PR acknowledges this. `unlockDeposit` is the riskiest to leave unexercised — its unlocking-period semantics are easy to get wrong and a static encode test can't catch a wrong message arg shape. Consider one live `unlockDeposit` → `getStackingInfo` assertion even if `withdraw` is skipped.

### 5. [Stacking] Base #302 is still OPEN

`feat/papi-2b-pallets` (#302) is unmerged; this PR stacks on it. Maintain merge order (302 → 303), or retarget 303 at `main` once 302 lands, to avoid a confusing diff in the GitHub UI (note: GitHub shows the full stacked diff including 2b's changes, which is why `.nvmrc`/`tests/package.json` appear in this PR's diff even though they're identical between 2b and 2c tips).

### 6. [Nit] Repo-wide Node 18 → 22 bump rides in this feature PR

The `.nvmrc` bump (needed for papi import-attributes support) and the `tests/package.json` test-script change land via 2c commits `90df46d`/`aa36d51`. Reasonable, but it's a toolchain change affecting every contributor and the CI matrix — worth a heads-up in the merge so it's not buried in a feature PR.

### 7. [Nit] `as any` casts (14 in `customers.ts`)

Consistent with the established 2a/2b pattern (`staking.ts` has 13) and documented (cross-runtime access for items absent from the mainnet static baseline). Acceptable as migration staging; becomes cleanup work at the 2d cutover when per-network descriptors are regenerated.

## Minor correctness notes

- `accountPlaceholder` (`5GrwvaEF…`) is canonical Alice (prefix 42) — fine for gas-sizing since `estimateGas` falls back to the generous ceiling when Alice can't afford `value`. Good.
- `isNoContract` via `getSs58AddressInfo` is robust to the all-zero AccountId32; the test's hardcoded Cere-prefix (54) zero SS58 is a fair mirror. Low risk since the storage type is fixed.
- `extractCreatedBucketIds`/`extractRemovedBucketIds` match the `Event` shape (`section`/`method`/`payload`) from `tx.ts`'s `shapeEvents` — correct.

---

**Recommendation:** Address #1 (verify the JSON resolves in the packed artifact) before merge — that's the only item that could ship a broken published build. The rest are follow-up-able.

_Review posted by ZCode._

spoluyan:
Follow-up verification of the review concerns (after the docs-cleanup + stack-linearization rebase):

**#1 — ABI JSON in the published artifact: not present (verified).** `resolveJsonModule: true` makes `tsc` emit `customer_deposit.json` into `dist/`. Confirmed by a clean rebuild (`rm -rf dist && tsc -p tsconfig.papi.json` → `dist/customer_deposit.json`, 41 KB) and `npm pack --dry-run` (listed under `dist`). The concern was reasonable but does not materialize with this tsconfig.

**#3 — owner consistency: fixed.** Applied the queried-`accountId` `owner` override to the `ClusterLedger` pallet fallback (parity with the contract path), and extended it to the legacy `Ledger` path added in 2d-iii. The entry is keyed by `accountId`, so `owner = accountId` is authoritative regardless of the decoded field.

**#2 readBalance owner override / #4 deposit-method coverage / #6 Node 18→22 bump / #7 `as any` casts:** intentional and documented (migration-guide note, funded-live-test follow-ups, migration staging). No change.

**`payload`/`shapeEvents`:** the #302 concern that `payload` does not deliver is a false positive — extractors read `payload.bucket_id`, which resolves (see the #302 thread).

Gates green post-rebase: check-types, lint, build (6 projects), 24 offline unit tests pass.

packages/blockchain/src/papi/pallets/customers.ts
```diff
@@ -0,0 +1,210 @@
...
+import type { Sendable, Event } from '../tx.js';
+import type { AccountId, Bucket, BucketId, BucketParams, ClusterId, StakingInfo } from '../../types.js';
+import { toBucket, toStakingInfo } from './mapping.js';
+import { createCustomerDepositContract, type CustomerDepositContract } from '../contracts/customerDeposit.js';
+
+// Placeholder caller for gas-sizing dry runs on payable/non-payable deposit
+// messages where the real signer is not yet known at build time (the dry run
+// only needs a valid AccountId, not funds — the estimateGas generous-ceiling
+// fallback covers the case where the placeholder can't afford `value`).
+const accountPlaceholder = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
```
yahortsaryk, line 11: 🟡 The `estimateGas` dry run is run as this placeholder for every contract-path build, but the placeholder can't afford `value` on the payable `deposit`/`depositFor`, and has nothing to unlock/withdraw on the non-payable ones — so the dry run fails at dispatch or contract level and `estimateGas` returns the `GENEROUS` ceiling anyway. Net: an extra RPC round-trip on every deposit/withdraw build whose precise result is almost always discarded. Not incorrect (ceiling is capped under `max_block`, unused gas refunded), but the sizing machinery is effectively dead here — consider short-circuiting to `GENEROUS` for these messages, or a comment on why the dry run is kept despite the placeholder.

packages/blockchain/src/papi/contracts/customerDeposit.ts
```diff
@@ -0,0 +1,162 @@
...
+      return { ...toStakingInfo(ledger), owner };
+    },
+
+    buildContractCall(contractAddr, message, value, args, gasLimit) {
+      const data = builder.buildMessage(message).call.enc(args ?? {});
+      return api.tx.Contracts.call({
+        dest: { type: 'Id', value: contractAddr } as any,
+        value,
+        gas_limit: gasLimit,
+        storage_deposit_limit: undefined,
```
yahortsaryk, line 162: 🟡 `storage_deposit_limit: undefined` means no cap — the signer implicitly covers whatever storage deposit the call incurs. Low risk for a deposit top-up, but an uncapped limit on a signed extrinsic is a footgun; consider a sensible cap.

tests/specs/papi-customers-deposit.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { connect } from '@cere-ddc-sdk/blockchain/papi';
+
+const describeChain = process.env.CERE_CHAIN_TESTS ? describe : describe.skip;
+
+// SS58 of the all-zero AccountId32 (Substrate default prefix) — the "no contract"
+// sentinel carried by a cluster with no deposit contract.
+const ZERO_SS58 = '5C4hrfjw9DjXZTzV3MwzrrAr9P1MJhSrvWGWqbYm2iZBonyC';
```
yahortsaryk: This sentinel is the wrong SS58 prefix. `customer_deposit_contract` decodes under Cere's prefix (54), but this is the substrate-default prefix-42 form (`5C4h…`), so `!== ZERO_SS58` never matches the chain's `6PWc…` zero address and the zero-cluster exclusion is a no-op — exactly the pitfall `papi-customers-write.spec.ts` documents and fixes with `ZERO_SS58_CERE`. Impact is mild here (falls back to `govEntries[0]`, and `getStackingInfo` handles zero via `isNoContract`), but the two specs are inconsistent — ideally export one shared sentinel/helper from the SDK and use it in both.

packages/blockchain/src/papi/pallets/customers.ts
```diff
@@ -0,0 +1,210 @@
...
+    unlockDeposit(clusterId, value) {
+      return contractOrPallet(
+        clusterId,
+        'DdcBalancesDepositor::unlock_deposit',
+        0n,
+        { value },
+        () => api.tx.DdcCustomers.unlock_deposit({ cluster_id: clusterId, value } as any) as Sendable,
+      );
+    },
+    withdrawUnlockedDeposit(clusterId) {
```
yahortsaryk, line 234: Contract-path live coverage is `deposit`/`depositExtra` only (acknowledged in the PR body). `depositFor`/`unlockDeposit`/`withdrawUnlockedDeposit` build via `as any` arg shapes (`{ owner }`, `{ value }`, `{}`) with no live assertion, so a wrong ink arg label would only surface at runtime. A cheap build+decode assertion (no submit) for these three would be good insurance given the contract path is the whole point.

packages/blockchain/src/papi/contracts/customerDeposit.ts
```diff
@@ -0,0 +1,162 @@
...
+  buildContractCall(contractAddr: string, message: string, value: bigint, args: any, gasLimit: Weight): Sendable;
+}
+
+export function createCustomerDepositContract(api: CereApi): CustomerDepositContract {
+  // ink dynamic builder (verified in the 2c spike): getInkLookup(abi) →
+  // getInkDynamicBuilder(lookup) exposes buildMessage(label) → { call, value },
+  // where `call.enc(args)` produces the SCALE input bytes (4-byte selector +
+  // encoded args) and `value.dec(bytes)` decodes the message return.
+  const builder = getInkDynamicBuilder(getInkLookup(abi as unknown as InkMetadata));
+  const cache = new Map<ClusterId, string | null>();
```
yahortsaryk, line 44: Nit: this cache never invalidates (same as the legacy `CustomerDepositContracts` in #300). Fine, just noting the "contract address is immutable for the client's lifetime" assumption.

yahortsaryk (commented):
Cleanest of the 2x series — no correctness blockers. I verified the bits most likely to be wrong (the Event extractors match the base's normalized `shapeEvents`, `hex()` passes SS58 strings through, the two-level ink decode, and all three specs match `papi*.spec.ts`) and they hold up. A couple of Medium notes + nits inline. The chain-reality comments are genuinely great.

packages/blockchain/src/papi/pallets/customers.ts
spoluyan: Confirmed against live chains — your analysis is exactly right, and I have direct evidence from the deposit debugging in #308.

The dry run for the payable `deposit`/`depositFor` messages runs as the unfunded placeholder *with `value`*, so it fails the balance check and `estimateGas` returns the `GENEROUS` ceiling every time. On a real devnet dry run the payable call came back `success=false` with `storage_deposit: Charge 0` and `gas_required: {ref_time: 0, proof_size: 0}` — so `gas_required` is not merely discarded, it's zero and would be unsafe to size from. The precise `2x gas_required` branch only ever runs for the non-payable messages.

Not a correctness bug (ceiling is capped under `max_block`, unused gas is refunded), but the round-trip is wasted on the primary path. The proper fix is to dry-run as the *real funded caller* rather than the placeholder — which also unblocks precise sizing for the payable messages — and that's tracked in #308, since the deposit builder needs to learn the caller for the storage-deposit-limit work anyway. Short-circuiting to `GENEROUS` would be the cheap alternative if we don't do the caller plumbing.

packages/blockchain/src/papi/contracts/customerDeposit.ts
spoluyan: Good catch, and it turned out to be more than a footgun — this line is a **live blocker on devnet**, filed as #308.

Verified by dry runs on both chains:

| Network | `storage_deposit_limit` | Result |
|---|---|---|
| devnet | `undefined` (None) | `StorageDepositLimitExhausted` — None is treated as a **0** limit |
| devnet | explicit 5 CERE | `StorageDepositNotEnoughFunds` (account underfunded) |
| testnet | `undefined` (None) | ✅ success, real charge **4.2 CERE** |
| testnet | explicit 5 CERE | ✅ success |

So devnet's `pallet-contracts` and testnet's disagree on what `None` means, and every **first-time** depositor on devnet fails (repeat deposits grow no storage, which is why it went unnoticed — both prior test accounts were already paired). Testnet accepts an explicit cap too, so "send a sensible cap" is the forward-compatible fix as you suggest; it needs the real funded caller for sizing (the placeholder can't afford a payable dry run — see the `estimateGas` thread). Parked in #308 pending the blockchain team's answer on which `None` semantics mainnet will land on.

tests/specs/papi-customers-deposit.spec.ts
spoluyan: Fixed — the deposit spec now uses the prefix-54 sentinel (`6PWc…`), matching `papi-customers-write.spec.ts`, with a comment explaining why the prefix-42 form silently no-ops. Your read was exactly right: it never matched, so the exclusion fell through to `govEntries[0]`.

Not exporting a shared helper from the SDK yet — it's test-only knowledge today and the two specs now agree; worth doing if a third consumer appears.

packages/blockchain/src/papi/pallets/customers.ts
spoluyan: Agreed, and live testing since made this concrete: `depositFor`/`unlockDeposit`/`withdrawUnlockedDeposit` still have no live contract-path assertion, and the deposit path turned out to have a real chain-level bug (#308) that only surfaced when actually submitting. Adding the cheap build+decode assertions for those three is a tracked follow-up; the funded-account work in #308 is the natural place since it needs the same setup.

packages/blockchain/src/papi/contracts/customerDeposit.ts
spoluyan: Noted — the assumption is that a cluster's deposit-contract address is immutable for the client's lifetime. A governance change would need a fresh client. Documented rather than invalidated, since `resolve()` is on the hot path for every deposit/balance call and clusters' contract addresses don't change in practice.

packages/blockchain/src/papi/contracts/customerDeposit.ts
spoluyan: Follow-up: the SDK-side fix for this is now unblocked in principle — see #308 for the full dry-run matrix. Noting here because this exact line (`storage_deposit_limit: undefined`) is the one that needs to change, so whoever picks up #308 should start here.

yahortsaryk:

The individual review threads here are resolved — the only thing blocking approval is the open deposit blocker #308 (`storage_deposit_limit=None` → every first-time deposit on a contract cluster fails on devnet), which you've confirmed is a live bug. The SDK-side fix lands right here (`customerDeposit.ts` `buildContractCall`), so this is the PR it's attached to.

Right now #308 is OPEN, unassigned, with no activity since it was filed (2026-07-23) and no linked fix PR, and it's gated on the blockchain team's answer about `None` semantics for mainnet.

1. Do you intend to land the #308 fix **before** merging this stack, or merge now and fix in a follow-up? If merge-then-fix, please state it explicitly in the PR description / MIGRATION.md so the devnet first-deposit breakage is a documented, known limitation of 3.0 rather than a surprise.

yahortsaryk (approved):
Approving. All review threads on this PR are resolved; the sole remaining hold was the #308 first-deposit blocker, now tracked for an SDK fix in #309 (the fix lands on this PR's `customerDeposit.ts` buildContractCall). Approving on the agreed merge-then-fix basis. Recommend noting the devnet first-deposit limitation in MIGRATION.md until #309 lands so it's a documented known issue.

PR #304, opened 2026-07-11
feat(ddc): Phase 2d-i — unified papi-native Signer (stacked on #303)

spoluyan:
## Phase 2d-i — Unified papi-native Signer (stacked on #303)

First of three sub-projects in the **2d consumer cutover** (2d-i signer → 2d-ii DdcClient/consumer cutover → 2d-iii legacy removal + new major + migration guide). Direction: a clean, papi-native, best-practices public API — breaking is acceptable under the new major.

Additive under `packages/blockchain/src/papi/signers/` (reworking the 2a signers), exposed via `@cere-ddc-sdk/blockchain/papi`. Base branch is `feat/papi-2c-customers` (#303). The legacy `src/Signer/` and the package `.` export are untouched (removed in 2d-iii).

### What's added
- **Chain-free `Signer` interface** matching the `@cef-ai/signer` seam byte-for-byte: `{ type: 'ed25519'|'sr25519'|'ecdsa'|'ethereum'; address; publicKey; isReady(); sign(bytes, intent?) }` + `isSigner`. A signer from either repo (or a CEF wallet) satisfies it structurally — the cross-repo seam just works. Replaces 2a's `CereSigner`.
- **`toPolkadotSigner(signer)`** — internal bridge adapting `sign(bytes, 'extrinsic')` to papi's `PolkadotSigner` (used by `client.tx.send`). Keeps the interface chain-free; callers still pass one signer object to both chain and storage. Throws for non-sr/ed types.
- **Five consolidated concrete signers** (papi-native): `KeyringSigner` (hdkd keypair base), `UriSigner` (seed/URI — absorbs 2a `MnemonicSigner`), `createRandomSigner`, `Web3Signer` (browser extension — absorbs 2a `ExtensionSigner`), `CereWalletSigner` (injected wallet), `JsonSigner` (encrypted keystore).

### Design
- Crypto is `@polkadot-labs/hdkd`(-helpers); **no `@polkadot/api`** in the signer layer. Only `JsonSigner` uses `@polkadot/util-crypto`/`@polkadot/keyring` (keystore decrypt).
- `JsonSigner` reads the curve from the keystore encoding and handles the sr25519 (full 64-byte key) vs ed25519 (32-byte seed) shape difference.

### Verification
- `sign()` output verifies against the public key for **both sr25519 and ed25519** (real `sr25519.verify`/`ed25519.verify` round-trips, not tautologies); `JsonSigner` decrypts deterministic sr25519 + ed25519 keystore fixtures (address matches `UriSigner`); wrong passphrase throws; `isSigner` rejects a raw `PolkadotSigner`; a real devnet `System.remark` signs+finalizes via `UriSigner` through the bridge.
- Full papi suite green: 13 suites / 40 tests (`--runInBand`); `check-types:ci`, `lint`, full `npm run build` all green. Existing papi specs migrated `MnemonicSigner`→`UriSigner`.

### Next (2d-ii / 2d-iii)
- **2d-ii:** DdcClient/cli/ddc/file-storage/BlockchainStrategy switch to the papi client + this `Signer` (storage layer imports the new `Signer`, passes `SignIntent`); remove the last legacy `Blockchain` usages.
- **2d-iii:** delete legacy `src/Signer/`, `Blockchain.ts`, `DDC*Pallet.ts`, `CustomerDepositContract.ts`, the `@polkadot/api` dep; promote the papi layer to the package `.` export (drop the `/papi` subpath); new major; migration guide (folding in dropped CDN methods, `StorageNodeMode.Compute`, the `Ledger.owner` root-cause, testnet/mainnet compat shim).


spoluyan:
## Review — Phase 2d-i unified papi-native Signer

**Verdict: ✅ Approve.** Clean, well-scoped design. I reproduced the verification claims independently (isolated worktree, `npm ci` from scratch) and traced the risky crypto/bridge paths against the installed `polkadot-api@2.1.8` source. Findings below are nits/observations, not blockers.

### Independently verified

| Gate | Result |
|---|---|
| `npm run build` (6 packages) | ✅ green |
| `npm run check-types:ci` | ✅ green |
| `npm run lint` | ✅ green |
| New signer unit specs (sr25519/ed25519 verify, JsonSigner decrypt, `isSigner`, bridge throws) | ✅ 13 passed / 1 skipped |
| Full papi suite serial (offline) | ✅ 13 passed / 27 skipped (devnet-gated self-skip) |
| No leftover refs to deleted `MnemonicSigner`/`ExtensionSigner`/`CereSigner` | ✅ clean |
| Node bump (`.nvmrc` → 22.11.0) propagates to all 3 CI workflows | ✅ they all read `.nvmrc` |

### The one architectural risk — confirmed safe

The `toPolkadotSigner` bridge passes `signer.sign(bytes, 'extrinsic')` (a **Promise**) where the old `MnemonicSigner` passed hdkd's **sync** `kp.sign`. I checked `@polkadot-api/signer`:

```ts
declare function getPolkadotSigner(publicKey, signingType,
  sign: (input: Uint8Array) => Promise<Uint8Array> | Uint8Array): PolkadotSigner;
```
and `from-raw-signer.js` does `const signed = await sign(...)` — async is supported and awaited. The bridge is correct.

### Observations (none blocking)

**1. `Web3Signer` extrinsic path changes extension UX semantics — flag for 2d-ii.**
The old `ExtensionSigner.getPolkadotSigner()` returned the extension's *native* signer (its own `signTx`, which drives PJS `signPayload` with `withSignedTransaction`). The new `Web3Signer` routes everything through `signer.sign()` → `account.polkadotSigner.signBytes()`. `signBytes` (in `@polkadot-api/signers-common`) wraps the payload in `<Bytes>…</Bytes>` and signs it as a *raw byte message*.

For the chain path, papi itself assembles `callData + signedExtensions`, calls the bridge's sign callback with that combined payload, then wraps the result in `createV4Tx` — so the signature is over the extrinsic payload and is **valid**. Not a correctness bug. But an extension/hardware wallet that distinguishes "sign transaction" vs "sign arbitrary bytes" in its consent UI will now always present the latter. That's a UX regression vs the old `ExtensionSigner`, and there's no unit test exercising a live extension. Worth confirming with a real extension before 2d-ii flips `DdcClient` onto this. (Scope-deferred here, so just track it.)

**2. `JsonSigner` sr25519 secret handling is correct.** Cross-checked `@polkadot/keyring/pair/index.js` (`createPair`): sr25519 uses the full 64-byte secret directly when `length === 64`; ed25519 expands from seed. The PR's comment + `secretKey.subarray(0, 32)` for ed25519 matches exactly — good that it's documented inline.

**3. Redundant eslint-disable comments (cosmetic).** The repo's eslint config sets `'@typescript-eslint/no-unused-vars': ['error', { args: 'none' }]` — function *arguments* are never flagged. So the 6× `// eslint-disable-next-line @typescript-eslint/no-unused-vars` above each `_intent?: SignIntent` param are dead (lint passes with or without them). Could trim for noise; harmless either way.

**4. Name collision with legacy signers is intentional.** `UriSigner`/`Web3Signer`/`CereWalletSigner`/`JsonSigner`/`KeyringSigner` now exist in **both** `.` (legacy) and `/papi`. Different import paths → no compile conflict; legacy set explicitly untouched (slated for 2d-iii). Anyone importing `@cere-ddc-sdk/blockchain` still gets the *old* `UriSigner` (with `getSigner()`); `/papi` gives the chain-free one. The 2d-ii design doc's structural-assignability note correctly explains why the cutover lands in green steps.

**5. CI cutover is sound.** Pointing `tests/package.json` `test` at `jest.papi.config.ts --runInBand` drops the flaky docker E2E + the CJS-ts-jest-vs-ESM-papi failure from CI. Papi specs self-skip without `CERE_CHAIN_TESTS`, so CI is deterministic/network-free; docker suite stays manually runnable. Same setup #306 uses.

### Recommendation

Merge as-is. Consider the `Web3Signer` extension-UX point (1) as a 2d-ii follow-up before the consumer cutover, and optionally trim the redundant eslint-disable comments (3).

---
*Verified in an isolated worktree off `feat/papi-2d-i-signer`; current working branch untouched.*

packages/blockchain/src/papi/signers/Web3Signer.ts
```diff
@@ -0,0 +1,57 @@
...
+    // InjectedPolkadotAccount.type is the optional keypair type ('ed25519' |
+    // 'sr25519' | 'ecdsa') the extension reports; default to sr25519 if absent.
+    return this.account.type ?? 'sr25519';
+  }
+  async isReady(): Promise<boolean> {
+    return true;
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async sign(bytes: Uint8Array, _intent?: SignIntent): Promise<Uint8Array> {
+    return this.account.polkadotSigner.signBytes(bytes);
```
yahortsaryk, line 40: 🟡 This makes browser-extension extrinsics sign through the extension's `signBytes`/`signRaw`, not its `signPayload`/`signTx`. When a `Web3Signer` goes through `client.tx.send`, `resolveSigner` sees `isSigner === true` and routes it via `toPolkadotSigner`, which re-wraps this `signBytes` callback — so extrinsic signing never reaches the account's native `signTx`.

Those paths aren't equivalent for pjs-style extensions:
- **Correctness:** extensions often handle `signRaw` differently from payload signing (some wrap the bytes), so the signature can fail to verify against the raw extrinsic payload → chain rejects it.
- **UX/security:** `signTx`/`signPayload` shows the user a decoded extrinsic in the prompt; `signBytes` shows opaque bytes. The old `ExtensionSigner` returned the account's native `polkadotSigner` (with `signTx`) directly and avoided both.

This path has no test (browser-only). Suggest special-casing `Web3Signer` for the `'extrinsic'` intent to use its native `signTx`, or verifying end-to-end against a real extension before relying on it. (This is where the `SignIntent` arg should earn its keep.)

packages/blockchain/src/papi/signers/JsonSigner.ts
```diff
@@ -0,0 +1,51 @@
...
+export class JsonSigner extends KeyringSigner {
+  constructor(json: KeystoreJson, passphrase: string) {
+    const { keypair, type } = decodeKeystore(json, passphrase);
+    super(keypair, type);
+  }
+}
+
+function decodeKeystore(json: KeystoreJson, passphrase: string): { keypair: KeyPair; type: 'sr25519' | 'ed25519' } {
+  // encoding.content is e.g. ['pkcs8', 'sr25519'] or ['pkcs8', 'ed25519'].
+  const type: 'sr25519' | 'ed25519' = json.encoding.content[1] === 'ed25519' ? 'ed25519' : 'sr25519';
```
yahortsaryk, line 30: Any non-`ed25519` keystore defaults to `sr25519` — an `ecdsa` keystore (valid in `@polkadot/keyring`) would silently decode/sign as sr25519 and produce garbage rather than failing. Consider throwing on an unsupported `encoding.content[1]` instead of defaulting.

packages/blockchain/src/papi/signers/JsonSigner.ts
```diff
@@ -0,0 +1,51 @@
+import { decodePair } from '@polkadot/keyring/pair/decode';
```
yahortsaryk, line 1: Nit: `@polkadot/keyring/pair/decode` is an internal subpath, not the package's public API, so it can break on a minor `@polkadot/keyring` bump with no semver signal. Worth a comment, or use the public surface if one exists.

packages/blockchain/src/papi/signers/types.ts
```diff
@@ -1,12 +1,33 @@
...
+  isReady(): Promise<boolean>;
+  sign(bytes: Uint8Array, intent?: SignIntent): Promise<Uint8Array>;
+}
+
+/** Structural guard: validates the full shape (chain-free — does not require any chain method). */
+export function isSigner(value: unknown): value is Signer {
+  if (!value || typeof value !== 'object') return false;
+  const c = value as Record<string, unknown>;
+  return (
+    typeof c.type === 'string' &&
```
yahortsaryk, line 27: Nit: this accepts any string `type` rather than the four valid `SignerType` values, so `{ type: 'nonsense', … }` passes the guard and only fails later in `toPolkadotSigner`. Restricting to the known set would fail faster and make `isSigner` a true validator.

yahortsaryk (commented):
Clean signer rework — the chain-free seam is a nice abstraction and the tests are genuinely good (real sr25519/ed25519 verify round-trips, wrong-passphrase check, isSigner rejecting a raw PolkadotSigner). One concern on the Web3Signer extrinsic path + a few nits inline.

packages/blockchain/src/papi/signers/Web3Signer.ts
spoluyan: This was correct and it bit exactly as predicted. Testing with the polkadot.js extension against devnet, `createBucket` failed with `InvalidTxError: BadProof` — the extension wraps `signRaw` bytes in `<Bytes>…</Bytes>`, so the reconstructed signature didn't verify against the raw extrinsic payload.

Fixed in 2d-iii: added a `NativePolkadotSigner` escape hatch — `Web3Signer` exposes `getPolkadotSigner()` returning the extension account's native `polkadotSigner`, and `toPolkadotSigner` prefers it when present. So extrinsics now go through the extension's `signTx`/`signPayload` (decoded prompt, correct signature) while data signing still uses `sign()`. Verified working on devnet and testnet with the real extension.

Your point that "this is where `SignIntent` should earn its keep" is the right framing — the intent is what lets the bridge pick the native path for extrinsics.

packages/blockchain/src/papi/signers/JsonSigner.ts
spoluyan: Fair — an `ecdsa` keystore would currently fall through to the sr25519 branch and produce garbage instead of failing. Not changed in this pass because no Cere wallet export path emits ecdsa (all are sr25519/ed25519), so it's a latent footgun rather than a live one. Throwing on an unrecognized `encoding.content[1]` is a small, safe improvement — tracked as a follow-up.

packages/blockchain/src/papi/signers/JsonSigner.ts
spoluyan: Acknowledged — `@polkadot/keyring/pair/decode` is an internal subpath with no semver guarantee. It's the only way to decrypt a keystore JSON without pulling the whole `Keyring` runtime, which is what the chain-free signer seam is trying to avoid. Worth the comment; if it breaks on a minor bump, the fallback is vendoring the ~30-line decode.

packages/blockchain/src/papi/signers/types.ts
spoluyan: Right that `isSigner` accepts any string `type` and defers the real failure to `toPolkadotSigner`. Deliberately structural: the guard exists to distinguish a chain-free `Signer` from a raw papi `PolkadotSigner` at the `tx.send` boundary, not to validate the curve. Narrowing it to the known set would make it a stricter validator and fail faster — reasonable, tracked as a follow-up rather than changed mid-stack since `resolveSigner` already rejects unsupported curves.

yahortsaryk:

yahortsaryk (approved):

PR #305, opened 2026-07-12
feat(ddc): Phase 2d-ii — DdcClient + consumer cutover to papi (stacked on #304)

spoluyan:
## Phase 2d-ii — DdcClient + consumer cutover (stacked on #304)

Second of three sub-projects in the 2d consumer cutover (2d-i signer ✅ → **2d-ii cutover** → 2d-iii legacy removal + new major + migration guide). Cuts the SDK's consumers off the legacy `@polkadot/api` `Blockchain` onto the papi client (2a–2c) + the chain-free `Signer` (2d-i). Base branch is `feat/papi-2d-i-signer` (#304). The legacy chain layer + `@polkadot/api` stay in place (deleted in 2d-iii).

### What changed
- **`@cere-ddc-sdk/blockchain/papi` is now the single consumer import source** — its barrel gained papi-native `encodeAddress`/`decodeAddress` (hdkd ss58), `UriSignerOptions`, and the domain types consumers use.
- **DdcClient** — `create(signer, config)` unchanged ergonomically; `config.blockchain` accepts a network name / ws-URL (→ papi `connect()` internally, DdcClient owns `disconnect()`) or an injected `CereClient`. All chain calls go through `client.customers`/`tx`/`chain`. The legacy `Blockchain`-instance option is removed.
- **FileStorage** — cut over the same way (it constructed its own `Blockchain` in standalone mode).
- **`Router`/`BlockchainStrategy`** take a `CereClient`; routing reads come from `client.nodes.listStorageNodes()` + `client.customers.getBucket()`.
- **`ddc` storage layer, public barrels, presets, cli, playground, specs** all import from `/papi`.

### Verification
- **Full workspace green** — `build` + `check-types` + `lint` across all 6 packages.
- **papi suite** 15 suites / 44 tests green, including a **devnet DdcClient chain-path e2e** (real `getBalance` + `getDeposit` through the papi-backed client).
- A grep gate confirms **no shipped consumer** (`ddc`/`ddc-client`/`file-storage`/`cli`/`playground`) imports the legacy `@cere-ddc-sdk/blockchain`. The only remaining legacy imports are four `tests/` files that test/bootstrap the legacy layer itself (removed in 2d-iii).

### Notes / green-build sequencing
- The one runtime hazard — a papi signer breaking the legacy `Blockchain.send`'s nominal `isSigner`/`instanceof` checks — was contained: `sdkToken`'s `Web3Signer` `instanceof` stayed on the legacy class until the barrel + all construction flipped to `/papi` together (end state one class).
- `BlockchainStrategy.isReady()` is memoized (resolve-once) to avoid a per-operation RPC round-trip.

### Deferred to 2d-iii
- Delete legacy `Blockchain`/`*Pallet`/`src/Signer/`/`CustomerDepositContract` + the `@polkadot/api` dep; migrate/remove the 4 legacy-layer test files.
- Promote `/papi` → the package `.` export (drop the `/papi` subpath); lift the per-package tsconfig `paths` shim to root; extract a shared `resolveClient()` helper.
- New major + migration guide (dropped CDN methods, `StorageNodeMode.Compute`, `Ledger.owner` root-cause, testnet/mainnet compat shim, playground signer-type polish, readyProbe rejection hardening).


spoluyan:
## PR Review — Phase 2d-ii Consumer Cutover to papi

Reviewed by reading the full diff + actual source in a throwaway worktree (didn't run the suites myself). The cutover is well-sequenced and the atomic-flip reasoning is sound. A few things worth attention before merge.

---

### 🔴 Should discuss before merge

**1. The `tests` package test-script change silently drops the entire integration suite from CI**

`tests/package.json` `test` went from `jest` → `jest --config jest.papi.config.ts --runInBand`. The papi config's `testMatch` is `specs/papi*.spec.ts` **only**. The default `jest.config.ts` matched `specs/**/*.spec.ts` **and** wired `globalSetup`/`globalTeardown` (the docker-compose chain + storage nodes).

So `npm test` / `npm run test:ci` (`lerna run test`) now **no longer runs**: `Auth`, `FileStorage`, `DdcClient`, `Errors`, `StorageNode`, `Routing`, `CacheControl`, `ActivityCapture`, `DdcApis`, `Blockchain`, `ChainCompat`, `metadata-audit`. The commit message (`c42a62c`) frames this as dropping "flaky docker-E2E + CJS run," but it's a large, quiet coverage reduction. Two things to decide explicitly:
- Is the docker suite intentionally demoted to **manual-only** (`jest --config jest.config.ts`)? If so, that deserves a line in the PR body / CONTRIBUTING, not just a commit message.
- Several of the dropped specs (`Auth`, `Errors`, `StorageNode`, `CacheControl`, `ActivityCapture`) are **not** docker-E2E and arguably should still run in CI. They're collaterally dropped because they share the default config.

**2. `BlockchainStrategy.readyProbe` caches the promise permanently — a transient first-op failure is unrecoverable**

`packages/ddc/src/routing/BlockchainStrategy.ts:31`:
```ts
this.readyProbe ??= this.probeLiveness();
return this.readyProbe;
```
The probe is one `getCurrentBlockNumber()` RPC. If that single call fails on the very first `getNode()` (transient WS drop, brief chain unavailability, a slow cold-connect timeout), the **rejected promise is memoized forever** — every subsequent store/read/dagNode op re-awaits the same rejection with no retry path. The old `api.isReady()` re-attempted. Suggested:
```ts
this.readyProbe ??= this.probeLiveness().catch((e) => { this.readyProbe = undefined; throw e; });
```

---

### 🟡 Worth noting

**3. Repo-wide changes bundled into a "consumer cutover" PR**
- `.nvmrc` v18.17.1 → v22.11.0 (justified — papi CLI uses `import … with { type: 'json' }`, needs Node ≥20.10). Affects every contributor and the whole CI matrix; not called out in the PR body.
- The `tests` test-script change (#1 above).

Both are legitimate, but they scope-creep beyond the stated phase and would be easier to bisect/revert as their own commits/PRs.

**4. `DdcClient.connect()` is now a no-op — readiness errors are deferred**
`connect()` just `return this;` (papi client connects lazily). A bad endpoint or down chain used to throw at `await DdcClient.create(...)`; now it surfaces at the first operation. Defensible (matches papi's model) but it's a behavioral change worth a migration-note line in 2d-iii.

**5. Stacked on an open chain**
Base is `feat/papi-2d-i-signer` (#304, OPEN ← #303 ← #306…). Can't merge independently — expected for this stack, just flagging review/merge ordering.

---

### 🟢 Verified correct / well done

- **`null` → `undefined` in `depositBalance`** is right: `getStackingInfo` returns `StakingInfo | undefined` (confirmed in `pallets/customers.ts`), so `currentDeposit === undefined` matches the actual contract.
- **The atomic-flip ordering is genuinely careful.** Commit `378056c` keeps `sdkToken`'s `Web3Signer instanceof` on the legacy class until the barrel + all construction flip together (`c5bc066`) — correctly avoiding the silent "wallet prompted on every op" hazard where a papi signer would fail a nominal `instanceof` check. Good catch, well documented.
- **The `signer.type as 'ed25519' | 'sr25519'` cast** (`sdkToken.ts`) is a justified type-narrowing — that path is only reached with ed/sr25519 (enforced in `createSignature`), and `createRandomSigner` is typed narrower than the chain-free `SignerType`. The comment is the safety net and it's accurate.
- **`/papi` barrel growth** (Task 1) is clean and additive; the new `papi-barrel.spec.ts` smoke test is a nice cheap guard.
- **`FileStorage` cutover** correctly fixes the latent bug (it was reading/spreading a `blockchain` field that Router renamed to `client`) and mirrors DdcClient's `ownsClient` lifecycle pattern. The discriminated-union constructor type (`{nodes} | {blockchain}`) is a real improvement.
- **`encodeAddress`/`decodeAddress`** via `@polkadot-labs/hdkd-helpers` ss58 correctly replaces `@polkadot/util-crypto` for the consumer surface.

---

### Minor / nits

- `createBucket` destructures `const [bucketId] = extractCreatedBucketIds(...)` and returns `undefined` silently if no `BucketCreated` event is found. **Pre-existing** behavior (the old code did the same), not a regression — but a spot to throw instead, someday.
- `ChainConfig = CereNetwork | string | CereClient` — `CereNetwork` is decorative here (subsumed by `string`); the real validation is the runtime `NETWORKS.includes()` check. The comment owns this honestly.
- Playground `toCereWalletSigner` hex-bridge: the `EmbedWallet.signMessage` string-in/string-out contract is honestly flagged as "not a guaranteed-stable wire format." Fine for a demo.

---

**Net:** The cutover itself is high quality — careful sequencing, good commit hygiene, correct type reasoning. Main ask: make the **CI coverage reduction (#1)** explicit and intentional before merge, and harden the **`readyProbe` memoization (#2)** against a poisoned cache.

packages/ddc/src/routing/BlockchainStrategy.ts
```diff
@@ -1,29 +1,42 @@
...
+    this.client = client;
   }
 
   async isReady() {
-    await this.blockchain.isReady();
+    // `Router.getNode()` calls `isReady()` on every store/read/dagNode operation.
+    // Memoize the probe so it runs at most once per instance and resolves
+    // instantly thereafter, mirroring the old polkadot.js `api.isReady`
+    // resolve-once semantics and avoiding a per-operation RPC round-trip.
+    this.readyProbe ??= this.probeLiveness();
```
yahortsaryk: 🟡 Memoizing the probe is a good idea, but this caches a **rejected** promise too: a transient RPC blip on the first store/read leaves `readyProbe` permanently rejected, so every subsequent `isReady()` on this instance fails forever — a regression from `api.isReady()`, which retried the connection internally.

This is also inconsistent with #303's `getMaxBlock`, which deliberately clears its memo on rejection. Same one-liner here:
```ts
this.readyProbe ??= this.probeLiveness().catch((err) => {
  this.readyProbe = undefined; // don't memoize a transient failure
  throw err;
});
```
The PR body defers "readyProbe rejection hardening" to 2d-iii, but since it's a one-liner and an availability bug (a blip wedges the routing instance), I'd fold it in here.

packages/ddc-client/src/DdcClient.ts
```diff
@@ -39,22 +58,34 @@ type DepositBalanceOptions = {
...
   constructor(uriOrSigner: Signer | string, config: DdcClientConfig = DEFAULT_PRESET) {
     const logger = createLogger('DdcClient', config);
-    const blockchain =
-      typeof config.blockchain === 'string' ? new Blockchain({ wsEndpoint: config.blockchain }) : config.blockchain;
+    const bc = config.blockchain;
+
+    let client: CereClient;
+    let ownsClient: boolean;
+
+    if (typeof bc === 'string') {
```
yahortsaryk: Nit: a string that isn't in `NETWORKS` is passed to `connect()` as a WS URL → `inferNetwork()` → unknown host defaults to **mainnet** (the footgun from #301). So `DdcClient.create(seed, { blockchain: 'devnett' })` (a typo) becomes a WS-connection attempt with mainnet descriptors rather than a clear error. Fixing `inferNetwork` to throw on an unrecognized host also cleans this up; alternatively validate the string here before handing it to `connect`. (Same pattern in `FileStorage`.)

yahortsaryk (commented):
Clean, consistent consumer cutover — verified the wiring typechecks (RouterConfig picks up `client` via BlockchainStrategyConfig) and the papi-native ss58 helpers keep the Cere prefix (54), so AuthToken/mapSignature output is unchanged. No blockers; one Medium + one nit inline.

packages/ddc/src/routing/BlockchainStrategy.ts
spoluyan: Folded in here as you suggested — `isReady()` now clears the memo on rejection so a transient blip re-probes instead of wedging the instance:

```ts
return (this.readyProbe ??= this.probeLiveness().catch((err) => {
  this.readyProbe = undefined;
  throw err;
}));
```

It was already fixed in 2d-iii (#306), but you're right that #305 should be correct standing alone, so it lands in both — same form, so the stacked rebase dedupes it.

packages/ddc-client/src/DdcClient.ts
spoluyan: Partly addressed: `DdcClient` now throws when `blockchain` is missing entirely (fixed in #307, with a unit test). The typo case you describe — `'devnett'` treated as a WS URL, then `inferNetwork()` defaulting an unknown host to mainnet — is still open and is really the #301 `inferNetwork` issue; see that thread for why I'd rather fix it at the source than validate in each consumer.

yahortsaryk:

yahortsaryk (approved):

PR #306, opened 2026-07-12
Phase 2d-iii: legacy @polkadot/api removal + /papi→. promotion + 3.0.0

spoluyan:
## Phase 2d-iii — Legacy Removal + `.`-Promotion + New Major (3.0.0)

The final sub-project of the `@polkadot/api` → `polkadot-api` (papi) migration. Deletes the dead legacy chain layer, makes the papi API the package's single `.` export, and cuts the new major. **Closes the papi migration (goals #1–#3).**

Stacked on #305 (2d-ii). Base: `feat/papi-2d-ii-consumer-cutover`.

### What changed
- **Deleted the entire legacy `@polkadot/api` chain layer** — `Blockchain`, the 5 `DDC*Pallet`s, `CustomerDepositContract`, `src/Signer/`, `surface.ts` + the compat/metadata-audit tooling, legacy `utils/`/`constants.ts`.
- **Purged the legacy deps** — `@polkadot/api`(+`api-base`/`api-contract`/`rpc-core`/`rpc-provider`/`types`/`types-codec`/`types-create`/`extension-dapp`/`extension-inject`/`util`) and `@cere/embed-wallet-inject`. Only `@polkadot/keyring` + `@polkadot/util-crypto` remain (JsonSigner keystore decrypt); the rest is papi + hdkd.
- **Promoted `/papi` → `.`** — the papi barrel is now `@cere-ddc-sdk/blockchain`. The `/papi` subpath, the four per-package tsconfig `paths` shims, and the root path entry are gone. **One import path.** ESM-only `.` export (papi tsc output; microbundle `build:node`/`build:web` dropped).
- **Decommissioned the docker-E2E + legacy-live-chain + surface-audit test suites** — the papi suite (`tests/specs/papi*.spec.ts`) is now the sole test suite.
- **Cleanups** — extracted a shared `resolveClient()` (de-dupes DdcClient/FileStorage construction, semantics preserved verbatim); made `BlockchainStrategy.readyProbe` retryable (no longer caches a rejected liveness probe); playground derives `CereWalletSigner.type` from the account.
- **`MIGRATION.md`** — 2.x → 3.0 guide (legacy→papi name map, before/after snippets, and the folded-in behavior changes: dropped CDN/`ddcStaking.serve`, `StorageNodeMode.Compute` gap, the `Ledger.owner` contract-decode override, per-network pallet↔contract deposit fallback, ESM-only). Package READMEs rewritten off legacy examples.
- **3.0.0** across the workspace (lerna + all SDK manifests + pinned interdeps); lockfile regenerated (79 legacy pkgs pruned).

### Verification
- ✅ `build` / `check-types:ci` / `check-types` / `lint` green across all packages.
- ✅ Full read/query/e2e-read papi suite green against live devnet (incl. the DdcClient e2e), all signer unit tests, and a clean standalone `tsc` build (confirmed the `customer_deposit.json` ABI ships at the correct relative path).
- ✅ Whole-branch review (opus): **Ready to merge = Yes** — no Critical/Important/Minor defects.
- ✅ Grep gates empty: no `@polkadot/api` in shipped SDK; no `/papi` subpath anywhere in source; no `2.16.1` in SDK manifests.

### ⚠️ Pre-publish checklist (open verification gap — not a code defect)
- The **4 transaction-submitting** papi suites (`papi-customers-write`, `papi-tx`, `papi.spec`, `papi-signers`) currently fail with `InvalidTxError: Payment` because the shared funded devnet account (`~/.cef/models-seed`) is **drained** after a long session of real funded writes. The version bump is byte-inert and these suites were green earlier in the migration. **Refund the devnet account and re-run** `CERE_CHAIN_TESTS=1 NODE_OPTIONS=--experimental-vm-modules npx jest --config jest.papi.config.ts --runInBand` to close this before publishing.

### Follow-ups (out of scope)
- `examples/cli` + `examples/node` stay at 2.16.1 with their own `@polkadot/api` (can't reference `@cere-ddc-sdk/* 3.0.0` until published); their example code is now stale against the removed legacy API — port to papi in a follow-up.
- Pre-existing WS-teardown race makes `papi-chain.spec.ts` exit nonzero with all assertions passing — teardown-hardening ticket.
- Orphaned `tests/fixtures/files/2.2mb.jpg` (was only used by a deleted spec) — optional cleanup.


spoluyan:
## Independent review — verified green ✅

I reviewed this in an isolated worktree (your `feat/single-cluster-config` branch was not touched). **Verdict: Approve with two minor cleanups.** I ran a clean `npm ci` and verified all four gates pass, confirmed the build ships artifacts at the export-mapped paths, and ran the offline papi specs.

### Verified green

| Gate | Result |
|---|---|
| `npm run build` (lerna, 6 projects) | ✅ passes |
| `npm run check-types` | ✅ passes |
| `npm run check-types:ci` (`tsc -p tsconfig.build.json`, includes `tests/`) | ✅ passes |
| `npm run lint` (`eslint --quiet`) | ✅ passes |
| Offline papi specs (signers / json / wallet / barrel / resolve-client) | ✅ 18 passed, 29 skipped (chain-gated) |
| `@polkadot/api` purged from all `packages/*/src` | ✅ only mentions are in comments |
| No `/papi` subpath import anywhere in source/tests | ✅ |
| No `2.16.1` remaining in any SDK manifest or lerna.json | ✅ |
| Build emits `dist/papi/index.js` + `dist/customer_deposit.json` at the paths the export map declares | ✅ |

The 4 funded-write suites I could not run are blocked by the drained devnet account — exactly as flagged in the PR body. That is an environment issue, not a code defect.

### Code quality — the good

- **`resolveClient`** — clean de-dup with correct `ownsClient` lifecycle semantics; `DdcClient.disconnect()` and `FileStorage.disconnect()` both guard on `ownsClient` before tearing down an externally-supplied client. Unit test covers all three input shapes.
- **`BlockchainStrategy.isReady()`** — correctly clears the memoized probe on rejection so a transient liveness failure does not permanently brick the instance.
- **`sdkToken` per-signer cache** — the `.catch` deletes a rejected promise so failures are never cached; directly addresses the wallet rate-limit problem.
- **`NativePolkadotSigner` bridge** — right call. `Web3Signer` now exposes the extension native signer for extrinsics instead of reconstructing a signature from `signBytes` (which would fail on-chain with `BadProof`).
- **Contract-first deposit fallback** — three-model runtime detection (ink contract / migrated `ClusterLedger` / legacy account-global `Ledger`) is intricate but well-commented per branch.

### Issues to address

**1. Root `package.json` still declares dead `@polkadot/api` + `@polkadot/util` deps (minor — the one real gap)**

`package.json` (root) is **not touched by this PR**, yet it still lists `@polkadot/api: ^16.4.9` and `@polkadot/util: ^13.5.7`. Nothing in `packages/*/src`, `scripts`, `playground/src`, or `tests/specs` imports these anymore (only `@polkadot/util-crypto` and `@polkadot/keyring` survive, and those live in `packages/blockchain/package.json`, not root). These are dead weight at the workspace root, and they are why the lockfile still resolves ~167 `@polkadot/api`-tree entries (all under nested `examples/*/node_modules` and root `node_modules`).

The PR body claims "79 legacy pkgs pruned" — that is the net reduction, but the root declarations leave the tree partially alive. This does not break anything (the per-package `dependencies` are authoritative for what ships), but it contradicts the stated purge goal and leaves future readers unsure whether `@polkadot/api` is still a real workspace dependency. Recommend dropping `@polkadot/api` and `@polkadot/util` from the root `package.json` and regenerating the lockfile in this PR or an immediate follow-up.

**2. PR description is stale re: examples (minor — doc accuracy)**

The PR body says:

> Follow-ups (out of scope): `examples/cli` + `examples/node` stay at 2.16.1 with their own `@polkadot/api` ... port to papi in a follow-up.

But commit `1af8ce5` ("port examples to papi 3.0 + post-migration cleanups") actually **did** port them: examples are at `3.0.0`, `8-sudo-batch-remove-buckets` uses `connect()`/`client.tx.sudoAs`, and examples 5/6 use the corrected bare-`passphrase` `JsonSigner` signature. There is no remaining `new Blockchain`, `.ddcCustomers`, or `@polkadot/api` anywhere in `examples/` or `playground/`. The description should reflect this — as written it understates the PR's completeness. (Purely a description-vs-reality mismatch; the code is correct.)

### Pre-merge notes (not blockers)

- The funded-account gap is real but as-described.
- `StorageNodeMode.Compute` enum gap is documented in `MIGRATION.md` and deferred — reasonable for a major-version cutover.
- `connect()` is synchronous (returns `CereClient`, not `Promise`) — papi connects lazily; consistent across `resolveClient`, `DdcClient.connect()`, example 8, and the migration guide.

The two cleanups are both low-risk and could go in either this PR or a fast-follow. The core migration — legacy deletion, dep purge, export consolidation, version bump, migration guide — is solid and verified green.

yahortsaryk (commented):
Migration finale is well-executed — and thanks for folding in the #304 (Web3Signer native signer) and #305 (readyProbe reject-clear) fixes. Two things I'd resolve before publishing 3.0.0, plus a coverage note:

**Coverage cliff (accept knowingly):** this deletes every storage-layer functional spec (`FileStorage`, `StorageNode`, `Auth`, `CacheControl`, `DdcApis`, `Errors`, `ActivityCapture`, `DdcClient`) with the docker harness, leaving only the chain-layer `papi*` suite. After 3.0.0 there's no automated coverage of the gRPC storage read/write/auth/ack/retry paths — the SDK's core. Not a behavior regression (that code is import-swap only), but a large permanent gap. Combined with the pre-publish note that the 4 tx-submitting suites are currently red (drained devnet account), the write path isn't green on this PR right now — I'd want the funded re-run + a plan to restore storage coverage before publish.

Inline: the ESM/CJS consumer mismatch (blocker) and the ledger-detection brittleness.

packages/blockchain/package.json
```diff
@@ -1,19 +1,12 @@
 {
   "name": "@cere-ddc-sdk/blockchain",
-  "version": "2.16.1",
+  "version": "3.0.0",
   "description": "Cere Blockchain client",
   "type": "module",
-  "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "dist/types/index.d.ts",
+  "main": "./dist/papi/index.js",
```
yahortsaryk, line 6: 🔴 Making `blockchain` ESM-only (dropping the CJS build / `require` condition) but leaving `ddc`, `ddc-client`, and `file-storage` on microbundle `--format esm,cjs` with `"require": "./dist/index.cjs"` breaks CJS consumers: microbundle externalizes this dep, so their emitted `dist/index.cjs` does `require("@cere-ddc-sdk/blockchain")` → now an ESM module → `ERR_REQUIRE_ESM` at runtime (repo `.nvmrc` floor is 22.11, before unflagged `require(esm)` in 22.12; older Node fails outright, and papi's tree risks `ERR_REQUIRE_ASYNC_MODULE` even where supported).

The papi jest suite only exercises the ESM path, so CI won't catch it, and the E2E suites that might have are deleted here. So 3.0.0 advertises `require` entry points it can't honor. Either make the consumers ESM-only too (drop their `require`/`.cjs`/`browser` microbundle outputs to match this package) or keep a CJS build of `blockchain`. If the intent is whole-SDK ESM-only, please confirm MIGRATION.md says so.

packages/blockchain/src/papi/pallets/customers.ts
```diff
@@ -37,6 +37,33 @@ export interface CustomersPallet {
...
+  // contract. It only matters on a runtime with no contracts (mainnet today) —
+  // on devnet/testnet every cluster has a contract, so this probe never runs.
+  // Detected by whether the account-keyed `Ledger` storage exists on the
+  // connected runtime (present only on the legacy runtime; the migrated runtime
+  // replaced it with `ClusterLedger`).
+  let accountKeyedLedger: boolean | undefined;
+  const isAccountKeyedLedger = async (): Promise<boolean> => {
+    if (accountKeyedLedger === undefined) {
+      try {
+        await api.query.DdcCustomers.Ledger.getValue(accountPlaceholder as any);
```
yahortsaryk: 🟡 Detecting the legacy pallet by string-matching the error against `/not found/i` is brittle. On a migrated runtime where `Ledger` isn't in the metadata, `api.query.DdcCustomers.Ledger` is `undefined`, so `.getValue` throws a `TypeError` ("Cannot read properties of undefined") — which doesn't match `/not found/i`, so it re-throws and breaks `getStackingInfo`/deposit for a no-contract cluster on a migrated runtime. Latent today (every devnet/testnet cluster has a contract, so this never runs), but fragile exactly where it'd eventually be needed. A metadata-existence check (`!!api.query.DdcCustomers.Ledger`) is more robust than matching an error string.

packages/blockchain/package.json
spoluyan: Confirmed and fixed. I reproduced it before changing anything: `require()` of the built `ddc-client` failed against the ESM-only `blockchain`.

Aligned the three consumer packages to ESM-only (dropped the `cjs` microbundle format and the `require` condition) — MIGRATION.md already declared the SDK ESM-only, so that was the documented intent; I've now made it explicit for these packages too.

One correction to the suggested scope: the `browser` condition is **kept**. `dist/browser.js` is a real artifact built from the web tsconfig, and dropping it breaks the playground — the node bundle's `stream/consumers` import can't resolve in a browser (verified: vite build failed with `Could not load stream-browserify/consumers`). So it's ESM-only, but still dual node/browser.

While verifying this I also found a **second, independent packaging break on the ESM path**: `tsc` emitted `import abi from '../../customer_deposit.json'` with no `with { type: 'json' }` attribute, so Node's ESM loader threw `ERR_IMPORT_ATTRIBUTE_MISSING` on any deposit/balance call. The ABI now ships as a generated `.ts` module (`npm run build:abi`). Plain-Node `import` of `ddc-client` now succeeds (21 exports, all 7 ink messages parsed) where it previously threw.

packages/blockchain/src/papi/pallets/customers.ts
spoluyan: Fixed as suggested — the metadata-existence check is the primary signal now:

```ts
const ledgerEntry = (api.query.DdcCustomers as any)?.Ledger;
if (ledgerEntry?.getValue == null) return (accountKeyedLedger = false);
```

The `/not found/i` match is kept only as a fallback for the case where the entry exists in metadata but has no value. Your reading of the failure mode was right: on a migrated runtime `.getValue` on `undefined` throws a `TypeError` that the string match would miss, re-throwing instead of yielding the negative signal.

spoluyan:
Both inline items are fixed, and verification since found a **second packaging break** on the ESM path that this PR would have shipped:

**1. ESM/CJS consumer mismatch (blocker) — fixed.** Reproduced first: `require()` of the built `ddc-client` failed against the ESM-only `blockchain`. `ddc`/`ddc-client`/`file-storage` are now ESM-only too (dropped the `cjs` format + `require` condition). One deviation from the suggestion: the `browser` condition is **kept** — `dist/browser.js` is built from the web tsconfig and the playground needs it (the node bundle’s `stream/consumers` import cannot resolve in a browser; verified by a failing vite build).

**2. NEW — ABI JSON import was broken on the ESM path too.** `tsc` emitted `import abi from "../../customer_deposit.json"` with no `with { type: "json" }` attribute, so Node’s ESM loader threw `ERR_IMPORT_ATTRIBUTE_MISSING` on *any* deposit/balance call in a plain-Node consumer. The ABI now ships as a generated `.ts` module (`npm run build:abi`). Plain-Node `import` of `ddc-client` now succeeds (21 exports; all 7 ink messages parse) where it previously threw. This also supersedes the earlier "#303 ABI verified fine" note — that check only confirmed the file was *present* in `dist`, not that it could be *imported*.

**3. Ledger-detection brittleness — fixed.** Metadata-existence check is now primary; the `/not found/i` match is a fallback only.

**Coverage cliff:** accepted knowingly, as you framed it. Restoring storage-path coverage without the docker harness is tracked as a follow-up.

**Write path:** now green — a full funded round-trip (deposit → createBucket → store a real file → read back byte-exact) passed on **testnet**. Mainnet chain writes work (bucket 1396 created) but its storage/CDN load balancers reject the grpc-web WebSocket upgrade with HTTP 403 (testnet/devnet return 200) — infra, not SDK. Devnet first-deposit is blocked by a runtime `storage_deposit_limit` divergence, filed as #308. The previously-red tx suites were the drained shared account.

Gates on the tip: build ✓, check-types ✓, lint ✓, 26 offline unit tests ✓.

yahortsaryk:

yahortsaryk (approved):

PR #307, opened 2026-07-22
Remove presets → single-cluster, dual-URL (storage/CDN) config

spoluyan:
## Remove presets → single-cluster, dual-URL config

Replaces preset-driven, multi-node client-side routing with a **single-cluster** SDK configured by **explicit URLs**. Stacked on #306 (papi 3.0). Breaking (lands in 3.0).

Spec: `docs/superpowers/specs/2026-07-22-remove-presets-single-cluster-design.md` · Plan: `docs/superpowers/plans/2026-07-22-remove-presets-single-cluster.md`

### What changed
- **Config:** `DdcClient.create(signer, { blockchain: network|wsUrl|CereClient, clusterId, storageUrl, cdnUrl? })`. No presets, no baked-in URLs (`MAINNET`/`TESTNET`/`DEVNET`/`*_STATIC`/`ConfigPreset`/`DEFAULT_PRESET` all deleted). `cdnUrl` defaults to `storageUrl`; missing `clusterId`/`storageUrl` throws.
- **Single cluster:** `clusterId` moved into config; cluster-scoped `DdcClient` methods (`createBucket`/`depositBalance`/`depositBalanceFor`/`getDeposit`/`unlockDeposit`/`withdrawUnlockedDeposit`) dropped their per-call `clusterId`. The lower-level `client.customers.*` pallet is unchanged (still takes `clusterId`); `packages/blockchain` untouched.
- **Routing → resolver:** deleted `BlockchainStrategy`/`StaticStrategy`/`NodeTypeStrategy`/`RandomStrategy`/`PingStrategy`/`BaseStrategy`/`Router`. A tiny `EndpointResolver` maps operations to endpoints — writes → `storageUrl`, reads → `cdnUrl ?? storageUrl`.
- **Transport by URL scheme:** `https://`/`wss://` → grpc-web `WebsocketTransport` (browser + Node); `grpc://IP` → native `GrpcTransport` (Node only). Browser path is `https://` only.
- **Retries preserved:** `BalancedNode` keeps exponential-backoff retry (configurable `retries`); with one endpoint it re-attempts the **same** URL (node-fallover removed). Covered by a new offline `papi-retry` unit test.
- **Consumers:** `cli`, `playground`, `examples/*` updated to the new config with the real endpoints — testnet `*.testnet.dragon-1.xyz`, devnet `*.devnet.dragon-1.xyz`, mainnet `*.dragon-1.xyz` (app-level network→URL maps; the library carries no URLs).

### Verification
- ✅ `build` / `check-types:ci` / `check-types` / `lint` across all packages; `examples/node/*` type-check clean.
- ✅ Grep gates empty: no presets/strategies/`Router`/baked URLs in `packages/*/src`.
- ✅ Unit tests: `EndpointResolver` (op→URL), config-validation (throws on missing config), retry-same-endpoint.
- ✅ **grpc-web-in-Node validated:** a headless `DdcClient.read` over `https://cdn.devnet.dragon-1.xyz` round-tripped in Node and returned the DAG node (live CDN) — the unified-`https://`-everywhere assumption holds for reads.
- papi suite green except 5 write-submit tests failing `InvalidTxError: Payment` — the shared **devnet funded account is drained** (environmental, unrelated to this branch).
- Whole-branch review (opus): **Ready to merge = Yes** (0 Critical/Important).

### Follow-ups (non-blocking)
- **Node grpc-web DX:** the Node `https://` path needs `globalThis.self` (and `WebSocket`; native on Node ≥21) — the SDK should auto-polyfill these for the Node grpc-web transport, or document them.
- **Write-path round-trip** not run end-to-end (needs a funded account) — the read path is validated; verify a store round-trip pre-publish.
- Pre-existing `FileStorage` `logErrors === false` inversion (not introduced here).


spoluyan:
## Code Review — verified locally

**Verdict: Approve.** The PR does exactly what its spec describes, the architecture is a clear improvement (~450 lines of `Router` + 6 strategies + ping/discovery replaced by a 111-line `EndpointResolver`), and every gate I re-ran is green against the actual PR contents.

### Re-ran locally (not just trusting the PR body)
- ✅ `check-types:ci` — clean
- ✅ `build` (all 6 packages) — clean
- ✅ `lint` — clean *(see note below about my checkout)*
- ✅ Offline papi specs — 6/6 pass (`papi-endpoint-resolver`, `papi-ddcclient-config`, `papi-retry`; the ERROR/WARN lines in output are the error logger recording the intentionally-injected transient errors in the retry test)
- ✅ Examples type-check — clean
- ✅ Grep gates empty: no `presets`/`ConfigPreset`/`DEFAULT_PRESET`/strategies/`new Router(`/baked URLs in `packages/*/src`
- ✅ No consumer still passes the old `clusterId` to `createBucket`/`depositBalance`/`getDeposit` etc.

> Note: `npm run lint` *initially* reported 3 errors, but they all came from two **untracked files in my local working tree** (`tests/specs/papi-zz-hiring-*.spec.ts` — `??` in `git status`, present on neither `HEAD` nor the base branch). Not part of the PR. With those stray files moved aside, lint is clean.

### Strengths
1. **`EndpointResolver` is genuinely tiny.** The operation→URL map (`READ_OPERATIONS` set, `STORE_*`→`storageUrl`, `READ_*`→`cdnUrl ?? storageUrl`) is exactly the right abstraction level.
2. **`transportForUrl` is clean** — scheme-driven selection, correct `ws://`→`http://` normalization for `WebsocketTransport`, sensible `Unsupported transport URL scheme` throw. The `GrpcTransport.web` stub now correctly takes `_options` and throws in the browser.
3. **`createResolverNode` helper** correctly de-duplicates the identical resolver wiring between `DdcClient` and `FileStorage`.
4. **Retry semantics preserved correctly.** `withRetry` is properly simplified: `getNode` is re-fetched per attempt (always the same endpoint), retryable gRPC codes rethrow, everything else `bail`s. `papi-retry.spec.ts` confirms both paths (2 transient failures → success on attempt 3; `INVALID_ARGUMENT` bails immediately). `NodeError extends RpcError`, so the `instanceof RpcError` check correctly catches the test's thrown `NodeError`.
5. **Config validation upfront** — throws in the constructor before touching the blockchain client, covered by unit test.
6. **SDK-token memoization** correctly lifted verbatim from `Router.getSdkToken`.

### Minor observations (non-blocking)

1. **`StorageNode.mode` is now vestigial.** `EndpointResolver` always passes `mode: StorageNodeMode.Full`, and `this.mode` is assigned but never read in `packages/ddc/src`. The `grpcUrl`/`httpUrl` fields are likewise only meaningful on the legacy `DefaultTransport` path. Fine to leave for this PR.
2. **Pre-existing typo carried forward:** `OpperationRetryOptions` (double-p). In the existing public API so renaming is a separate breaking change — not this PR's job.
3. **`FileStorage` config-validation asymmetry.** `DdcClient` throws clearly when `clusterId`/`storageUrl` are missing, but `FileStorage` doesn't validate `storageUrl` — an undefined `storageUrl` flows into `EndpointResolver` and surfaces as a confusing transport error later. A small `if (!storageUrl) throw` would match `DdcClient`'s defensiveness. Optional.
4. **CLI default-URL map duplicates the playground's** (`networkToEndpoints` vs `NETWORK_ENDPOINTS`) with identical `*.dragon-1.xyz` hostnames. Acceptable per the spec ("app-level maps; the library carries no URLs"), but they'll drift in two places if endpoints change. Flagging for awareness.

### Risk & follow-ups
Low. Breaking change explicitly landing in the 3.0 major already in flight. The listed follow-ups are all real and correctly scoped out of this PR.

**Recommendation: merge** once the **write-path round-trip** is confirmed against a funded account — the one follow-up that gates real confidence in the unified-`https://`-everywhere assumption for *writes* (reads are already validated per the PR body). The 5 `InvalidTxError: Payment` papi failures are from the drained shared seed, not this branch.

packages/ddc/src/transports/transportForUrl.ts
```diff
@@ -0,0 +1,47 @@
...
+    const ssl = SSL_SCHEME.test(url);
+
+    /**
+     * `WebsocketTransport` only understands `http(s)://` hosts (it derives the
+     * actual `ws`/`wss` protocol internally from the `ssl` flag), so normalize
+     * a `ws://`/`wss://` endpoint URL to its `http(s)://` equivalent first.
+     */
+    const httpUrl = url.replace(/^wss:\/\//i, 'https://').replace(/^ws:\/\//i, 'http://');
+
+    return new WebsocketTransport({ httpUrl, ssl, ...options });
```
yahortsaryk, line 43: 🟡 This routes both browser and Node `https://` through grpc-web `WebsocketTransport`, which (per your follow-up note) needs `globalThis.self` and `WebSocket` in Node — without them a plain Node consumer hits `ReferenceError: self is not defined`. Since "unified `https://` everywhere" is the primary recommended Node path, I'd not ship this silently: the passing headless read test almost certainly ran in an env that already had those globals (jest / Node ≥21 WebSocket), masking the gap for a plain `node script.js`.

Suggest auto-polyfilling in the Node transport path (`globalThis.self ??= globalThis`, provide `WebSocket` where missing) rather than pushing it onto every consumer — or at minimum document it prominently in README/MIGRATION, not just a PR follow-up.

packages/ddc-client/src/DdcClient.ts
```diff
@@ -53,22 +71,41 @@ export class DdcClient {
...
+      throw new Error('DdcClient config is missing required "clusterId"');
+    }
+
+    if (!config?.storageUrl) {
+      throw new Error('DdcClient config is missing required "storageUrl"');
+    }
 
-  constructor(uriOrSigner: Signer | string, config: DdcClientConfig = DEFAULT_PRESET) {
     const logger = createLogger('DdcClient', config);
     const { client, ownsClient } = resolveClient(config.blockchain);
```
yahortsaryk, line 93: Nit: `blockchain` isn't validated the way `clusterId`/`storageUrl` are above. If it's omitted, `resolveClient(undefined)` returns `{ client: undefined }` and the first chain call (deposit/bucket/getBalance) crashes with an opaque error. A matching guard would keep the config failure modes consistent.

packages/ddc/src/routing/EndpointResolver.ts
```diff
@@ -0,0 +1,111 @@
...
+   * @returns A promise that resolves to the `StorageNode` for the resolved endpoint.
+   */
+  async getNode(operation: RouterOperation, bucketId: BucketId, config: Partial<StorageNodeConfig> = {}) {
+    const url = resolveUrl(operation, this.storageUrl, this.cdnUrl);
+
+    this.logger.info('Resolving endpoint for operation "%s" in bucket %s', operation, bucketId);
+
+    const authToken = await this.getSdkToken();
+
+    const storageNode = new StorageNode(this.signer, {
```
yahortsaryk, line 92: Optional: this builds a fresh `StorageNode` + transport on every operation and every retry attempt. It's a carried-forward pattern (the old `Router.getNode` did the same, so not a regression), but with only 1–2 static endpoints it's now trivially cacheable — memoizing the `StorageNode` per resolved URL would avoid re-instantiating the transport on each read/write.

yahortsaryk (commented):
Clean, well-scoped refactor — `EndpointResolver` is a much simpler replacement for the strategy stack, `transportForUrl`'s scheme handling is correct, and the retry-same-endpoint unit test pins down the right behavior. No blockers; one Medium (Node grpc-web polyfills) + two nits inline.

packages/ddc/src/transports/transportForUrl.ts
spoluyan: I tested this specifically because your reasoning about the spike environment was fair — but the failure doesn't reproduce on the supported Node floor, so I'd rather not add the polyfill.

Ran a full `DdcClient.read` against the live devnet CDN in **plain `node`** (no jest, no test harness):

```
typeof self: undefined | WebSocket: function
PLAIN-NODE READ OK: true
```

`self` is indeed undefined, but the grpc-web WebSocket transport doesn't need it — it needs a global `WebSocket`, which Node provides from 22, and the repo's `.nvmrc` floor is 22.11. I also drove `grpc.WebsocketTransport()` directly with `self` undefined and the socket opened fine.

So rather than polyfill something that isn't missing, I took your fallback suggestion and documented the real requirement in MIGRATION.md (new "Node version" subsection): Node ≥ 22.11, because the storage transport relies on a global `WebSocket`; on older Node the storage paths fail even though chain calls work. If we ever drop the floor below 22, the polyfill becomes necessary and this is the note to revisit.

packages/ddc-client/src/DdcClient.ts
spoluyan: Fixed — `blockchain` is now guarded alongside `clusterId`/`storageUrl`. Also applied the same treatment to `FileStorage`, which wasn't validating `storageUrl` at all despite its own docstring calling it required (flagged in the other review too). Both have unit tests now (offline suite 24 → 26).

packages/ddc/src/routing/EndpointResolver.ts
spoluyan: Agreed it's cheaply cacheable, and you've correctly noted it isn't a regression (the old `Router.getNode` rebuilt per call too). Leaving it as-is for this PR: with 1–2 static endpoints the cost is a small object plus a lazily-connecting transport, and the retry path deliberately re-resolves per attempt — memoizing would need care to avoid caching a transport whose socket has gone away. Tracking as a follow-up rather than folding it into this refactor.

yahortsaryk:

yahortsaryk (approved):

Issue #309, opened 2026-07-27
SDK follow-up: send an explicit storage_deposit_limit on ink! deposit calls (unblocks devnet first-time deposits)


spoluyan:
Follow-up to #308, split out so the papi migration stack (#300–#307) can merge without another 6-branch rebase cascade.

## Scope

Stop sending `storage_deposit_limit: undefined` (None) on the ink! deposit/withdraw contract calls; send an explicit, sized cap instead.

**Location:** `packages/blockchain/src/papi/contracts/customerDeposit.ts` → `buildContractCall`:

```ts
return api.tx.Contracts.call({
  dest: { type: 'Id', value: contractAddr } as any,
  value,
  gas_limit: gasLimit,
  storage_deposit_limit: undefined,   // <-- this
  data,
} as any) as Sendable;
```

## Why it matters

Devnet's `pallet-contracts` reads `None` as a **zero** limit, so **every first-time deposit on devnet fails** with `Contracts.StorageDepositLimitExhausted`. Testnet reads `None` as unlimited and works. Same SDK bytes, opposite runtime semantics — full dry-run matrix in #308.

Not a blocker for the stack (devnet-only; testnet verified end-to-end including a fresh account's first deposit; mainnet has no deposit contract and takes the legacy pallet path), but it is a **devnet DX regression vs 2.x**: `main` has no contract code at all and calls `ddcCustomers.deposit` directly, so 2.x never hits this path on devnet. A new developer onboarding on devnet currently cannot make their first deposit.

## Proposed fix

Send an explicit cap. Evidence that this works on both runtimes (from #308's dry runs):

| Network | `storage_deposit_limit` | Result |
| --- | --- | --- |
| devnet | `None` | `StorageDepositLimitExhausted` (None → 0) |
| devnet | explicit 5 CERE | `StorageDepositNotEnoughFunds` — *account underfunded (6.2 CERE < 4.2 storage + 2 value)*, not a rejection of the cap |
| testnet | `None` | ✅ success, real charge **4.2 CERE** |
| testnet | explicit 5 CERE | ✅ success |

Note (correcting an earlier assumption): this may **not** require plumbing the real funded caller through `DdcClient → customers → customerDeposit`. An explicit cap succeeded on testnet without it, and devnet's explicit-cap failure was a funding shortfall, not a cap rejection. Prefer the smallest change that works:

[...]

Issue #308, opened 2026-07-23
First-time deposit fails on devnet: storage_deposit_limit=None treated as 0 by pallet-contracts


spoluyan:
## Summary

A brand-new customer account cannot make its **first** DDC deposit through the SDK on any contract-bearing cluster on **devnet**. The deposit transaction fails with `Contracts.StorageDepositLimitExhausted`. The identical operation succeeds on **testnet**, because the two runtimes interpret an unset `storage_deposit_limit` (`None`) differently.

This was found during live cross-network verification of the papi 3.0 SDK (funded ed25519 account `6US5V6cQotopMThpmgaPBQzLK8JxjZ7yCcXCZkKhpNbvYq2d`).

## Impact

- **Devnet: every first-time depositor is blocked** on a contract cluster (all devnet/testnet clusters use the ink! deposit contract). Repeat deposits to an already-paired account work (no storage growth), which is why this went unnoticed — both prior test accounts were already paired.
- **Testnet: works** (full deposit → createBucket → store → read round-trip verified).
- **Mainnet: unaffected** — mainnet uses the legacy pallet `DdcCustomers.Ledger` deposit path, not the ink! contract.

## Root cause (proven by live dry-runs)

The ink! contract deposit call is submitted with `storage_deposit_limit = None` (`packages/blockchain/src/papi/contracts/customerDeposit.ts`, `buildContractCall` → `api.tx.Contracts.call({ ..., storage_deposit_limit: undefined })`).

- **Devnet's `pallet-contracts` treats `None` as a zero limit** → any storage-growing call (a first deposit creates a new ledger record) is rejected with `StorageDepositLimitExhausted`.
- **Testnet's runtime treats `None` as unlimited** → the same call succeeds.

Dry-run evidence (ink message `DdcBalancesDepositor::deposit`, via `ContractsApi.call`):

| Network | `storage_deposit_limit` | Result |
|---------|-------------------------|--------|
| devnet  | `None`            | `StorageDepositLimitExhausted` (limit treated as 0) |
| devnet  | explicit 5 CERE   | `StorageDepositNotEnoughFunds` (test account underfunded, see below) |
| testnet | `None`            | ✅ success — actual storage charge **4.2 CERE** |
| testnet | explicit 5 CERE   | ✅ success |

Reproduction (dry-run, devnet):
```js
await client.api.apis.ContractsApi.call(
  caller, contractAddr, 20_000_000_000n /* 2 CERE value */,
[...]

PR #310, opened 2026-07-28
chore(release): pre-3.0.0 cleanup — stale docs, dead scripts, engines floor

spoluyan:
Follow-up sweep after the papi stack (#300–#307) landed, before 3.0.0 is cut. **No runtime or source behavior changes** — docs, CI config, scripts and package metadata only.

## Docs that would otherwise ship wrong to npm

- **`ddc-client` / `file-storage` READMEs still documented the removed `TESTNET` preset** (`DdcClient.create(seed, TESTNET)`, `FileStorage.create(seed, TESTNET)`). These are the npm landing pages for 3.0.0, so a new user's first copy-paste would not compile. Rewritten to the explicit `blockchain` / `clusterId` / `storageUrl` config, with an ESM-only + Node ≥22.11 note linking MIGRATION.md.
  - Also fixes two latent bugs in the same snippet: `createBucket` no longer takes a cluster argument (it uses `config.clusterId`), and step 4 referenced an undefined `fileUri`.
- **`npm run docs` was broken on `main`.** `packages/blockchain/typedoc.json` still pointed at `tsconfig.web.json`, deleted when the package went ESM-only. Repointed at `tsconfig.papi.json` + `src/papi/index.ts` (`src/` also holds the generated ABI, which must stay out of the docs).
- **Regenerated the checked-in API docs for all four packages.** They still described `Blockchain`, `DDCCustomersPallet`, `DDCStakingPallet`, `CustomerDepositContracts` and `MnemonicSigner` — none of which exist in 3.0. Hence the large diff: ~5.7k lines of docs for deleted classes removed, 21 pages added for the papi surface (`connect`, `resolveClient`, `isSigner`, `inferNetwork`, …).

## CI

The test step carried `continue-on-error: true`, justified by a docker storage-node E2E that no longer exists — meaning **a real test regression could not fail the job**. Now blocking. The remaining suite runs offline and skips live-chain specs unless `CERE_CHAIN_TESTS` is set, so it stays deterministic in CI.

## Dead plumbing

- `test:env` / `test:clean` root scripts and the `test:env` nx target — no package has implemented them since `tests/setup` was removed.
- Root README documented a docker-based local environment that is gone. Replaced with the real gating (`CERE_CHAIN_TESTS`, `CERE_FUNDED_SEED`, devnet-only) plus a warning that the write-path specs submit signed extrinsics and spend real tokens.

## Packaging

`"engines": { "node": ">=22.11" }` added to all five publishable packages. Only `cli` declared one, stale at `>=18`. MIGRATION.md already documents the floor (the storage transport needs a global `WebSocket`); without `engines`, older-Node consumers hit an opaque transport failure instead of an npm warning at install. Verified it survives `clean-publish` into `package/`.

## Verification

| Gate | Result |
| --- | --- |
| `npm run build` | 0 |
| `npm run lint` | 0 |
| `npm run test:ci` | 0 — 26 passed, 29 chain-gated skipped |
| `npm run docs` | 0 (was failing before this PR) |
| `npm run package` | 0, `engines` present in output |

## Note for whoever cuts the release

`npm run release` (`lerna version --conventional-graduate`) will compute **3.1.0**, not 3.0.0: `lerna.json` is fixed-mode at `3.0.0`, and the only conventional commit since the last tag is the squashed `feat(ddc): Phase 1 protocol compatibility …` — the `!` / `BREAKING CHANGE:` footers from the branch commits did not survive squash-merging. Recommend `npx lerna version 3.0.0` explicitly (`exact: true` is already set), then hand-writing the breaking-change section pointing at MIGRATION.md.

Release notes should also carry a known-issues line for #308 / #309 (devnet first-time deposit fails; testnet and mainnet unaffected).


yahortsaryk (approved):
Approving. Docs/CI/metadata only, no source behavior changes, and it resolves two earlier review findings: the test step drops `continue-on-error` (my #300 concern — safe now that the suite is offline/deterministic with live specs CERE_CHAIN_TESTS-gated), and the `engines: >=22.11` floor turns the #307 Node WebSocket requirement into an install-time warning instead of an opaque runtime failure. Also fixes a broken `npm run docs`, removes a duplicated README paragraph, and corrects the ddc-client/file-storage README examples that still referenced the removed TESTNET preset (verified store() returns a FileUri, so the fixed snippet compiles). Only nit: the regenerated docs predate #311's { from } DepositOptions param — a post-merge `npm run docs` once both land will pick it up (already acknowledged in the PR description).

PR #311, opened 2026-07-28
fix(blockchain): send an explicit storage_deposit_limit on ink! deposit calls

### LINKED ISSUE #309: SDK follow-up: send an explicit storage_deposit_limit on ink! deposit calls (unblocks devnet first-time deposits)

spoluyan:
Fixes #309. Also answers two of #308's open questions with measurements (details below).

## The bug

Devnet's `pallet-contracts` reads an unset `storage_deposit_limit` (`None`) as a **zero** ceiling, so every storage-growing call — i.e. every **first-time** customer deposit — was rejected with `Contracts.StorageDepositLimitExhausted`. Testnet reads `None` as unlimited, so the identical SDK bytes worked there. `buildContractCall` now always sends an explicit limit.

## Why the limit is derived rather than a constant

Overshooting is not free. The runtime charges **`value + limit`** against the caller's spendable balance and rejects with `StorageDepositNotEnoughFunds` when that exceeds it — *even when the actual charge would have been affordable*. Measured live on devnet: `limit = 5 CERE` + `value = 1 CERE` is rejected against a 6.19 CERE balance, while the real charge is only 4.2. A large constant ceiling would therefore have traded one failure mode for another, blocking deposits that the account could actually afford.

## What changed

`estimateGas` → `sizeCall`, returning gas **and** the storage ceiling from the same dry run. This also resolves the dead branch flagged in review on #300/#303: the old code always dry-ran as an unfunded placeholder, so a payable deposit always failed and always fell back to the `GENEROUS` gas ceiling (devnet returned `gas_required: 0/0` — unusable, not merely discarded).

Sizing probes in two steps:

1. Dry-run with `None`. On a runtime that reads it as unlimited this succeeds and reports the true charge, keeping the common case to a single round trip.
2. If that fails, retry under the caller's **affordable headroom** (free balance less `value` and the existential deposit). The probe is affordable by construction, which is what gets the run far enough to report the charge on a `None`-means-zero runtime.

The limit is then `charge + existentialDeposit` — one ED of slack for state drift, deliberately tight. If both probes fail (unfunded caller, reverting contract, unhealthy node) it falls back to `10 × ED`: still explicit, never `None`.

Because those dry runs are only meaningful against the real account, the five deposit/withdraw builders take an optional `{ from }` (`DepositOptions`), and `DdcClient` passes its own signer address. Omitting it preserves the previous conservative behavior, so this is backward compatible.

## Live verification

Funded ed25519 account, devnet + testnet:

| Case | Limit sent | Result |
| --- | --- | --- |
| testnet — genuine new ledger record (`depositFor` to a fresh owner) | **5.2 CERE** (4.2 measured charge + 1 ED) | ✅ block `0x458a99e3…`, record created with `active = 1 CERE`, payer spent **5.23 CERE** — derived limit matched the real cost |
| testnet — repeat deposit, no storage growth | **1 CERE** (charge 0 + 1 ED) | ✅ `active` increased by exactly the deposited amount; no regression |
| devnet — first deposit | 10 CERE (fallback; caller unfunded so both probes fail) | failure changed from `StorageDepositLimitExhausted` → `StorageDepositNotEnoughFunds` |

That last row is the point: the **`None`-semantics defect is gone**. What remains on devnet is the test account being genuinely underfunded — 6.19 CERE against ~4.2 storage + value + ED + fees — which is the top-up requested in #308, not an SDK issue. A real devnet first-deposit submission can be confirmed once that lands; nothing in the SDK blocks it.

## Findings for #308

Two of its questions are now answered by measurement rather than opinion:

- **The 4.2 CERE charge is just the configured pricing**: `DepositPerItem 0.15 × 2 + DepositPerByte 0.06 × ~65 bytes = 4.2`. Not a devnet anomaly.
- **Those constants and the existential deposit are identical on devnet and testnet** (ED = 1 CERE, decimals = 10). The divergence is purely in how `None` is interpreted, not in how storage is priced.

Consequently the SDK no longer depends on which semantics mainnet converges to — an explicit limit is accepted by both. #308's remaining value is the runtime-intent question and the devnet top-up.

## Tests

8 new offline tests (`tests/specs/papi-customers-sizing.spec.ts`) drive the sizing logic against a fake `ContractsApi` that reproduces **both** runtime semantics, including the `value + limit` affordability rule observed live. They cover: an explicit limit is always sent; the limit equals `charge + ED`; the limit stays affordable; recovery via the second probe on a `None`-means-zero runtime; only one dry run when `None` is accepted; the all-probes-failed fallback; a failed balance read; and a `Refund`-only call.

| Gate | Result |
| --- | --- |
| `npm run build` | 0 |
| `npm run lint` | 0 |
| `npm run test:ci` | 0 — **34 passed** (26 before), 29 chain-gated skipped |

## Note on ordering

Independent of #310 (pre-release cleanup) — no overlapping files. #310 regenerates the checked-in API docs; the new optional parameter will be picked up by a `npm run docs` run after both land, so docs are deliberately not regenerated here to avoid a conflicting diff.

### DIFF
--- a/packages/blockchain/src/papi/contracts/customerDeposit.ts
+++ b/packages/blockchain/src/papi/contracts/customerDeposit.ts
@@ -24,15 +24,35 @@ type Weight = { ref_time: bigint; proof_size: bigint };
 // ~277M ref_time / ~53KB proof_size, far below this.
 const GENEROUS: Weight = { ref_time: 10_000_000_000n, proof_size: 1_000_000n };
 
+// Last-resort `storage_deposit_limit`, as a multiple of the chain's existential
+// deposit, used only when every sizing dry run failed (so the real charge is
+// unknown). It must be non-zero — see `sizeCall` for why `None` is never sent —
+// but it must also stay small, because the limit counts against the caller's
+// spendable balance: the runtime rejects `value + limit` above what the account
+// can afford with `StorageDepositNotEnoughFunds`, even when the *actual* charge
+// would have been affordable. Measured live: a new customer ledger record costs
+// 4.2 CERE (2 storage items + ~65 bytes at the chain's DepositPerItem/PerByte),
+// and the existential deposit is 1 CERE on devnet/testnet — so 10x ED clears a
+// first deposit with >2x headroom without needlessly inflating the balance the
+// caller must hold.
+const FALLBACK_SDL_IN_ED = 10n;
+
+/** Gas + storage-deposit ceiling for one contract call, sized by dry run. */
+export type CallSizing = {
+  gasLimit: Weight;
+  /** Explicit `storage_deposit_limit`. Never `undefined` — see `sizeCall`. */
+  storageDepositLimit: bigint;
+};
+
 export interface CustomerDepositContract {
   /** Contract address for a cluster (from gov params), or undefined when none/zero. Cached. */
   resolve(clusterId: ClusterId): Promise<string | undefined>;
   /** A customer's balance from the contract, or undefined when the account has no deposit. Throws on a failed dry run. */
   readBalance(contractAddr: string, owner: AccountId): Promise<StakingInfo | undefined>;
-  /** Dry-run a message to size its gas (2x margin, capped at the chain ceiling); falls back to the generous ceiling on failure. */
-  estimateGas(contractAddr: string, message: string, caller: AccountId, value: bigint, args: any): Promise<Weight>;
+  /** Dry-run a message to size both its gas and its storage-deposit ceiling. Never throws. */
+  sizeCall(contractAddr: string, message: string, caller: AccountId, value: bigint, args: any): Promise<CallSizing>;
   /** Build a signed-ready `Contracts.call` extrinsic for a contract message. */
-  buildContractCall(contractAddr: string, message: string, value: bigint, args: any, gasLimit: Weight): Sendable;
+  buildContractCall(contractAddr: string, message: string, value: bigint, args: any, sizing: CallSizing): Sendable;
 }
 
 export function createCustomerDepositContract(api: CereApi): CustomerDepositContract {
@@ -56,7 +76,7 @@ export function createCustomerDepositContract(api: CereApi): CustomerDepositCont
         .catch((err) => {
           // Don't memoize a rejection — a transient fetch failure would
           // otherwise permanently break `cap()` (and thus every `cap()`
-          // caller, including `estimateGas`) for the lifetime of this
+          // caller, including `sizeCall`) for the lifetime of this
           // contract instance. Clear the memo so the next call retries.
           maxBlock = undefined;
           throw err;
@@ -82,6 +102,42 @@ export function createCustomerDepositContract(api: CereApi): CustomerDepositCont
     };
   };
 
+  // Existential deposit, used as the unit for storage-deposit headroom. Cached
+  // like `getMaxBlock` (constants are a papi METHOD returning a Promise), and
+  // likewise never memoizing a rejection.
+  let existentialDeposit: Promise<bigint> | undefined;
+  const getExistentialDeposit = () => {
+    if (!existentialDeposit) {
+      existentialDeposit = api.constants.Balances.ExistentialDeposit()
+        .then((ed: any) => BigInt(ed))
+        .catch((err) => {
+          existentialDeposit = undefined;
+          throw err;
+        });
+    }
+    return existentialDeposit;
+  };
+
+  // What the caller could put toward a storage deposit for this call: free
+  // balance less the value being transferred and the existential deposit it must
+  // retain. Used as the probe limit on runtimes that reject `None` (see
+  // `sizeCall`) — by construction the probe is affordable, which is what lets the
+  // dry run get far enough to report the real charge. Returns 0n when the caller
+  // has no headroom, or on any read failure (the caller then falls through to the
+  // constant fallback rather than propagating).
+  const affordableHeadroom = async (caller: AccountId, value: bigint): Promise<bigint> => {
+    try {
+      const [account, ed] = await Promise.all([
+        api.query.System.Account.getValue(caller as any) as Promise<any>,
+        getExistentialDeposit(),
+      ]);
+      const headroom = BigInt(account.data.free) - value - ed;
+      return headroom > 0n ? headroom : 0n;
+    } catch {
+      return 0n;
+    }
+  };
+
   return {
     async resolve(clusterId) {
       const cached = cache.get(clusterId);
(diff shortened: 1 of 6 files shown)


yahortsaryk (dismissed):
_(Duplicate review — double-posted by tooling and dismissed. See the approving review just below for the actual approval + inline notes.)_

packages/blockchain/src/papi/contracts/customerDeposit.ts
```diff
@@ -93,34 +149,78 @@ export function createCustomerDepositContract(api: CereApi): CustomerDepositCont
...
+      const gr = sized.gasRequired;
+      return {
+        // 2x safety margin — the dry run is priced against possibly-stale state,
+        // so a verbatim `gas_required` can undershoot.
+        gasLimit: await cap({ ref_time: BigInt(gr.ref_time) * 2n, proof_size: BigInt(gr.proof_size) * 2n }),
+        // The measured charge plus one existential deposit of slack for state
+        // drift between this dry run and the real call. Deliberately tight rather
+        // than generous: the limit is checked against the caller's spendable
+        // balance, so padding it would reject deposits the account can afford.
+        storageDepositLimit: sized.charge + ed,
```
yahortsaryk: 🟡 Edge case: the `charge + ed` slack isn't guaranteed affordable. The headroom probe succeeds when `value + charge + ed ≤ free` (probe limit is `headroom = free − value − ed`, runtime checks `value + limit ≤ free − ed`), but the returned limit is `charge + ed`, so the real submit needs `value + charge + 2·ed ≤ free`. For a caller in the window `free − 2·ed < value + charge ≤ free − ed`, sizing *succeeds* yet the actual tx is rejected with `StorageDepositNotEnoughFunds` — the drift-slack pushes a tight-but-affordable deposit over the spendable line.

Honest failure, narrow (~1 ED) window, not a blocker. But since the probe already proved affordability at `headroom`, capping the result at `min(charge + ed, headroom)` would ensure a run the probe proved affordable can't then be rejected. (The boundary isn't covered by the tests — the tight-limit test gives +5 CERE of slack.)

packages/blockchain/src/papi/contracts/customerDeposit.ts
```diff
@@ -93,34 +149,78 @@ export function createCustomerDepositContract(api: CereApi): CustomerDepositCont
...
+      if (!sized) {
+        const headroom = await affordableHeadroom(caller, value);
+        if (headroom > 0n) sized = await attempt(headroom);
       }
-      const gr = dry.gas_required;
-      // 2x safety margin — the dry run is priced against possibly-stale state and
-      // a placeholder caller, so a verbatim `gas_required` can undershoot.
-      return cap({ ref_time: BigInt(gr.ref_time) * 2n, proof_size: BigInt(gr.proof_size) * 2n });
+
+      const ed = await getExistentialDeposit().catch(() => 0n);
```
yahortsaryk: 🟡 `getExistentialDeposit().catch(() => 0n)` can collapse the limit to zero: if the ED read fails, `ed = 0n`, so the fallback becomes `FALLBACK_SDL_IN_ED * ed = 0` (line 210) and the success-path limit becomes `charge + 0` (line 222). An explicit `storage_deposit_limit: 0` reintroduces the exact `StorageDepositLimitExhausted` this PR fixes on a `None`-means-zero runtime (and drops all drift slack) — despite the `FALLBACK_SDL_IN_ED` comment saying it "must be non-zero." ED is a chain constant so this is rare, but the defensive `catch` backfires; a non-zero floor (or not sizing off a zeroed ED) would keep the "never effectively zero" guarantee.

yahortsaryk (approved):
Approving. This solves the original #308/#309 blocker: `buildContractCall` now always sends an explicit `storage_deposit_limit` instead of `None`, so the devnet `None`-means-zero → `StorageDepositLimitExhausted` first-deposit failure is gone. It also cleanly resolves two earlier review findings — the `estimateGas` run-as-unfunded-placeholder dead-path (now priced against the real caller via `{ from }`) and the MIGRATION doc. The `None`-first / affordable-headroom two-step probe is a nice minimal-round-trip design, and the offline harness that reproduces both runtime semantics + the `value + limit` affordability rule is exactly right.

Two edge-case robustness notes inline — neither is a blocker, both are fine as small follow-ups.

packages/blockchain/src/papi/contracts/customerDeposit.ts
spoluyan: Good catch, and confirmed real — pushed a fix in `52aad06`.

The defensive `catch` backfired exactly as you describe: with `ed = 0n` the fallback became `FALLBACK_SDL_IN_ED * 0 = 0`, i.e. `storage_deposit_limit: 0`, which on devnet is the `StorageDepositLimitExhausted` this PR exists to kill. `getExistentialDeposit` now defaults to the Cere ED (1 CERE — the value on mainnet/testnet/devnet) on a persistent read failure instead of rejecting, rather than letting callers swallow to `0n`. ED is a chain constant, so a read failure means the node is unhealthy, not that ED is zero; this mirrors `getChainDecimals`' fallback to `10` for the same reason. The memo still clears on failure so transient errors self-heal.

Covered by two new tests: the fallback stays non-zero when `ExistentialDeposit` throws, and the success path keeps its `charge + default ED` slack instead of collapsing to `charge + 0`.

packages/blockchain/src/papi/contracts/customerDeposit.ts
spoluyan: Verified the algebra — this is real, and your `min(charge + ed, headroom)` fix is the right one. Pushed in `52aad06`.

The probe proves affordable at `charge` (it runs with limit `headroom = free - value - ed`, and the runtime's check `value + limit <= free - ed` is then a tautology), but the returned `charge + ed` needs `value + charge + 2*ed <= free` — one ED tighter. So in the window `free - 2*ed < value + charge <= free - ed` sizing succeeds yet the submit is rejected with `StorageDepositNotEnoughFunds`. Capping the result at `headroom` closes it, and since a successful probe implies `charge <= headroom`, the cap stays `>= charge` so it can't swing the failure back to `LimitExhausted`.

`headroom` is now computed once up front so the cap applies on both runtimes (the None-means-unlimited path doesn't otherwise prove affordability); it's one extra `System.Account` read, no extra dry run. Added a test for the tight window: free 6.5 CERE → uncapped limit would be 5.2 (rejected), capped limit is 4.5, which still covers the 4.2 charge and fits within spendable.

PR #312, opened 2026-07-30
fix(ci): refresh package-lock.json and guard against staleness

spoluyan:
Unblocks the 3.0.0 release. The publish run ([30500853727](https://github.com/Cerebellum-Network/cere-ddc-sdk-js/actions/runs/30500853727/job/90739982262)) failed at the final step:

```
lerna ERR! EUNCOMMIT Working tree has uncommitted changes, please commit or remove the following changes before continuing:
lerna ERR! EUNCOMMIT  M package-lock.json
```

`lerna publish from-package` refuses to publish from a dirty tree. **Nothing was published** — it aborted before touching the registry. The tree was dirtied two steps earlier by `npm ci --legacy-peer-deps`, which rewrites `package-lock.json` whenever it disagrees with the workspace `package.json` files.

Reproduced with a pristine `git clone` of `main` plus CI's exact install command.

## Two independent sources of drift

**1. Missing `engines` (introduced by #310).** Five `engines: { node: ">=22.11" }` blocks were added to the publishable packages but never mirrored into the lockfile — it carries each workspace's metadata, so it went stale immediately:

```
packages/blockchain:  None              -> {"node": ">=22.11"}
packages/cli:         {"node":">=18.0.0"} -> {"node": ">=22.11"}
packages/ddc, ddc-client, file-storage: None -> {"node": ">=22.11"}
```

**2. 242 stale entries (pre-existing, not from #310).** 195 under `examples/*` plus 47 hoisted `node_modules/*` still pinned `@cere-ddc-sdk/* 2.16.1`/`2.15.0` **tarballs from the registry**. Those workspaces now declare `3.0.0`, which resolves to the local workspace packages instead, so `npm ci` drops all 242. Confirmed by reproducing on `421373e` (pre-#310): also dirty, same 242 entries. **This would have failed the release on its own.**

Regenerating the lockfile fixes both: `4648 insertions, 8541 deletions`.

## The guard

Adds `git diff --exit-code package-lock.json` immediately after `npm ci` in **both** workflows:

- `publish.yaml` — fails fast with an actionable message instead of an opaque `EUNCOMMIT` at the last step.
- `tests.yaml` — catches the PR that *introduces* the drift. None of `build`/`lint`/`check-types`/`test` run `npm ci`, which is exactly why this slipped through #310's gates.

## Verification

| Check | Result |
| --- | --- |
| `npm run build` | 0 |
| `npm run lint` | 0 |
| `npm run test:ci` | 0 — 37 passed, 29 chain-gated skipped |
| `npm run package` | 0 |
| `npm ci` in a **pristine clone** → tree clean | ✅ (was `M package-lock.json`) |
| guard step passes | ✅ |

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -49,6 +49,19 @@ jobs:
       - name: Install packages
         run: npm ci --legacy-peer-deps
 
+      # `npm ci` rewrites package-lock.json when it disagrees with the workspace
+      # package.json files. That leaves the tree dirty, and `lerna publish` then
+      # aborts with an opaque `EUNCOMMIT ... M package-lock.json` at the very last
+      # step (run 30500853727). Fail here instead, with a message that says what to do.
+      - name: Verify package-lock.json is up to date
+        run: |
+          if ! git diff --exit-code --quiet package-lock.json; then
+            echo "::error::package-lock.json is out of date — 'npm ci' rewrote it."
+            echo "Run 'npm ci --legacy-peer-deps' locally in a clean checkout and commit the result."
+            git diff --stat package-lock.json
+            exit 1
+          fi
+
       - name: Build packages
         run: npm run build
 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -39,6 +39,19 @@ jobs:
       - name: Install packages
         run: npm ci
 
+      # Catch a stale lockfile on the PR that introduces it, rather than at the
+      # publish step where it surfaces as `lerna ERR! EUNCOMMIT` (run 30500853727).
+      # Adding/changing `engines`, deps, or a workspace's version all require the
+      # lockfile to be regenerated and committed alongside.
+      - name: Verify package-lock.json is up to date
+        run: |
+          if ! git diff --exit-code --quiet package-lock.json; then
+            echo "::error::package-lock.json is out of date — 'npm ci' rewrote it."
+            echo "Run 'npm ci' locally in a clean checkout and commit the result."
+            git diff --stat package-lock.json
+            exit 1
+          fi
+
       - name: Check code style
         run: npm run lint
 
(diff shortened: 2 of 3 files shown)


spoluyan:
## Blocked on review — this PR gates the 3.0.0 release

`main`'s branch protection requires 1 code-owner approval, so I can't self-merge (and shouldn't `--admin` past it). Publish stays broken until this lands.

One note for the reviewer, since it looks alarming: **CI failed on the first push and that was the new guard working correctly.** `tests.yaml` ran plain `npm ci` while `publish.yaml` runs `npm ci --legacy-peer-deps`; the two modes record ~26 transitive packages as `dev` vs `peer`, so no single lockfile could satisfy both and each workflow would reject the other's. Fixed in `33613ae` by aligning `tests.yaml` to `--legacy-peer-deps` (matching the release path rather than changing resolution for publishing). Green now.

## Two further findings from dry-running the rest of the release

Both verified in a throwaway clone; **neither is in this PR**, they're for the release step that follows:

1. **`lerna version` corrupts the lockfile.** It drops the `node_modules/@polkadot-api/descriptors` workspace link and marks entries `extraneous`/`devOptional`. Committing lerna's lockfile would reintroduce this exact `EUNCOMMIT` failure. The version commit must run `npm ci --legacy-peer-deps` *after* `lerna version` and commit that result — confirmed a stable fixpoint over 3 consecutive runs.

2. **`npm run release` would publish 3.1.0, not 3.0.0.** `lerna.json` is fixed-mode at 3.0.0 and the squash-merged commits carry no `BREAKING CHANGE:` footer, so `--conventional-graduate` computes a minor bump — skipping 3.0.0 entirely. Confirmed by dry run: `npx lerna version 3.0.0` produces correct CHANGELOGs with the `v2.14.1...v3.0.0` compare link. The breaking-change section has to be hand-written (conventional-changelog can't derive it from those squashes).

Full runbook prepared; I'll execute it and trigger Publish as soon as this is approved and merged.

PR #314, opened 2026-07-30
docs: fix UriSigner examples that throw, and overstated CommonJS restriction

spoluyan:
Two documentation defects that shipped in 3.0.0, found by installing the published packages from npm and actually running the documented snippets. **No code changes.**

## 1. Every quick-start example throws

```ts
const signer = new UriSigner('//Alice');   // Error: UriSigner: empty mnemonic/URI
```

3.0's `UriSigner` deliberately refuses to fall back to a built-in development phrase — a missing seed should fail loudly rather than silently sign with a well-known key. That's correct behaviour, but the docs were never updated, so **the first line a new user copies fails**.

Fixed in all 8 places: `MIGRATION.md` ×5, `packages/blockchain/README.md`, and the `DdcClient.create` / `FileStorage.create` TSDoc — the last two matter most because they ship *inside the published tarballs* and surface in editor tooltips.

The replacement prefixes the standard development mnemonic, which I verified derives the **canonical Alice key** (`0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d`) against the published package — so the examples are accurate, not merely non-throwing. Added a short note in MIGRATION.md explaining why the bare path no longer works.

## 2. The CommonJS restriction is overstated

MIGRATION.md and the CHANGELOG said `require()` "is not supported in 3.0". I tested it: 3.0 requires Node ≥ 22.11, Node 22 loads ESM from `require()`, and `require('@cere-ddc-sdk/blockchain')` from a CommonJS project **succeeds**.

Reworded to what's actually true: no CommonJS build is shipped, `require()` may work via Node's interop, don't depend on it (bundlers and other runtimes are free to reject it). The old wording would push people into refactors for a restriction that doesn't bite them.

## Also

- **`playground.yaml` aligned** to `npm ci --legacy-peer-deps`, matching `publish.yaml`/`tests.yaml`. Plain `npm ci` there rewrites the lockfile — harmless for a deploy (no `lerna publish`), but it left the three workflows disagreeing about the lockfile, the same inconsistency #312 fixed. Caught while verifying the playground build.
- **API docs regenerated**, which also picks up `CallSizing` and `DepositOptions` from #311.

## Verification

Every corrected snippet executed against the **published 3.0.0** from npm, not the repo:

```
1. UriSigner example      OK -> 6UJs2Uj5hV2Jto7Z2FAhM2hXpvvQtcr4dwM6abKzMYQBHmpi
2. encode/decodeAddress   OK
3. getAccountFreeBalance  OK   (live testnet)
4. customers.deposit      OK   (explicit storage_deposit_limit sent)
5. FileStorage.create     OK
```

| Gate | Result |
| --- | --- |
| `npm run docs` | 0 |
| `npm run lint` | 0 |
| `npm run build` | 0 |
| `npm run test:ci` | 0 — 37 passed, 29 chain-gated skipped |

Worth a `3.0.1` patch afterwards so the corrected TSDoc reaches npm consumers — the README/TSDoc fixes only help people reading GitHub until republished.

### DIFF
--- a/.github/workflows/playground.yaml
+++ b/.github/workflows/playground.yaml
@@ -34,7 +34,7 @@ jobs:
           node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
 
       - name: Setup GitHub Pages
-        uses: actions/configure-pages@v3
+        uses: actions/configure-pages@v5
 
       # Must match publish.yaml/tests.yaml. Without --legacy-peer-deps npm records
       # some transitive packages as `peer` rather than `dev` and rewrites the
@@ -52,4 +52,4 @@ jobs:
 
       - name: Deploy
         id: deployment
-        uses: actions/deploy-pages@v2
+        uses: actions/deploy-pages@v4
--- a/.github/workflows/playground.yaml
+++ b/.github/workflows/playground.yaml
@@ -36,8 +36,11 @@ jobs:
       - name: Setup GitHub Pages
         uses: actions/configure-pages@v3
 
+      # Must match publish.yaml/tests.yaml. Without --legacy-peer-deps npm records
+      # some transitive packages as `peer` rather than `dev` and rewrites the
+      # lockfile, so the three workflows would disagree about its contents.
       - name: Install packages
-        run: npm ci
+        run: npm ci --legacy-peer-deps
 
       - name: Build packages
         run: npm run build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,10 @@ upgrade guide, with before/after examples for every renamed API, is in
   `client.nodes`, `client.customers`, `client.staking`, `client.clustersGov`, and
   extrinsic submission at `client.tx.send(tx, { signer })`.
 * **packaging:** `blockchain`, `ddc`, `ddc-client` and `file-storage` are
-  **ESM-only** — they no longer publish a CommonJS entry, so `require()` is not
-  supported. Use `import`, a dynamic `await import(...)`, or a bundler.
+  **ESM-only** — they no longer publish a CommonJS entry. Use `import`, a dynamic
+  `await import(...)`, or a bundler. (Node 22 can load ESM from `require()`, so a
+  `require()` call may happen to work — that is Node interop, not a supported
+  entry point.)
 * **node:** **Node >= 22.11** is now required (declared via `engines`). The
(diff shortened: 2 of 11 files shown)


spoluyan:
Added a third fix (`cc395a7`) after the release-triggered playground deploy failed — folding it in here since it's the same workflow file this PR already touches.

## The playground deploy is broken (pre-existing, not a 3.0.0 regression)

[Run 30542210185](https://github.com/Cerebellum-Network/cere-ddc-sdk-js/actions/runs/30542210185) — build and upload succeeded, the **Deploy** step failed:

```
##[error]Getting signed artifact URL failed
##[error]HttpError: Cannot find any run with github.run_id 30542210185
Failed to create deployment (status: 404) ... Ensure GitHub Pages has been enabled
```

That last hint is a **red herring** — Pages is enabled and set to `build_type: workflow` (verified via the API). The real cause is a version mismatch: `upload-pages-artifact@v3` stores the artifact on the new artifact backend, and `deploy-pages@v2` can't read it — v2 predates that backend and calls the retired artifact API, which is why it 404s looking up its own run id.

Bumped `deploy-pages` v2 → v4 (the version paired with upload v3) and `configure-pages` v3 → v5. `upload-pages-artifact` stays at v3.

Pre-existing: the workflow last succeeded in January, before `upload-pages-artifact` went to v3 — so nothing in the 3.0.0 release caused it.

Once this merges, the deploy needs a re-run (`workflow_dispatch`), since the `release: published` event has already fired for v3.0.0. I'll trigger it and confirm the playground is live.

yahortsaryk (approved):
Approving. Docs + one CI-config line, no code changes, and both fixes are real 3.0.0 defects verified against the published npm package: the throwing `UriSigner('//Alice')` examples (the doc side of the empty-phrase behavior flagged on #301 — replacement correctly derives the canonical Alice key) and the overstated "require() not supported" wording (now accurately hedged as Node ESM interop, don't rely on it). Fixing the DdcClient/FileStorage TSDoc matters most since it ships in the tarball. Two minor notes, neither blocking: the playground.yaml action bumps (configure-pages v3→v5, deploy-pages v2→v4) aren't called out in the description; and `require(esm)` is unflagged only from Node 22.12 (behind a flag on exactly 22.11.0) — the hedged wording stays correct regardless. Agree a 3.0.1 patch afterward gets the TSDoc fixes to npm consumers.

